PR #20, opened 2021-08-02
feat: helpers for signing blobs

coryan:


gcs/iam.py
```diff
@@ -0,0 +1,49 @@
...
+import binascii
+import flask
+import json
+
+import testbench
+
+IAM_HANDLER_PATH = "/iamapi"
+iam = flask.Flask(__name__)
+iam.debug = False
+iam.register_error_handler(Exception, testbench.error.RestException.handler)
```
cojenco, line 27: Q: Since the registering handlers are moved to their own files, I’m assuming we no longer need them in `emulator.py`?

tests/test_iam.py
```diff
@@ -0,0 +1,58 @@
...
+        path, app = gcs.iam.get_iam_app()
+        self.client = app.test_client()
+
+    def test_sign_blob(self):
+        response = self.client.post(
+            "/projects/-/serviceAccounts/test-service-account:signBlob",
+            data=json.dumps(
+                {
+                    "payload": base64.b64encode(
+                        "please sign this".encode("utf-8")
```
cojenco, line 37: :)

gcs/iam.py
coryan: Yes, I think when I move the `emulator.py` file I will remove the additional registrations.  Now that I know how to unit test Flask applications that might get better testing too.

cojenco (approved):

Issue #24, opened 2021-08-03
Copy the existing code from google-cloud-cpp


coryan:
I probably should have created this earlier.

- [x] Copy holder.py and its test
- [x] Copy bucket.py and its test
- [x] Copy object.py and its test
- [x] Copy database.py and create a test for it
- [x] Copy emulator.py and create a test for it, maybe rename to testbench.py
- [ ] Copy grpc_server.py and create a test for it

coryan:
We decided not to move the storage/v1 gRPC implementation. I can now close this issue.

PR #26, opened 2021-08-05
feat: implement Object resource

coryan:
Part of the work for #24.

This will probably conflict with #25 in the `__init__.py` file, meh.


coryan:
This is too large, I will break it down.

PR #30, opened 2021-08-08
feat: add initial resource database

coryan:
testbench.database will hold all the resources (Bucket, Object, etc.) so
the Flask and gRPC applications can find them and implement the various
RPCs associated with them.

Part of the work for #24

### DIFF
--- a/testbench/database.py
+++ b/testbench/database.py
@@ -1,0 +1,246 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import collections
+import json
+import os
+import uuid
+import re
+
+import gcs
+import testbench
+
+from schema import Schema, SchemaError, And, Use
+from google.cloud.storage_v1.proto import storage_pb2 as storage_pb2
+
+
+class Database:
+    def __init__(
+        self,
+        buckets,
+        objects,
+        live_generations,
+        uploads,
+        rewrites,
+        retry_tests,
+        supported_methods,
+    ):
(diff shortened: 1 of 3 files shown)


coryan:
Ping

testbench/database.py
```diff
@@ -0,0 +1,228 @@
...
+            metageneration, match, not_match, True, context
+        )
+
+    def get_bucket_without_generation(self, bucket_name, context):
+        bucket = self.buckets.get(bucket_name)
+        if bucket is None:
+            testbench.error.notfound("Bucket %s" % bucket_name, context)
+        return bucket
+
+    def insert_bucket(self, request, bucket, context):
```
cojenco, line 64: How can `metageneration` be applied to the bucket when calling `insert_bucket`? Comparing this to `insert_test_bucket` where on line 104
```
bucket_test.metadata.metageneration = 4
bucket_test.metadata.versioning.enabled = True
```

Context: When creating a new bucket against the emulator (ultimately calling `insert_bucket`, the new bucket does not  have a `metageneration` attached. Is this WAI? If so, how can one add a `metageneration` to a newly created bucket?

cojenco (commented):
A question on `insert_bucket`. Otherwise, LGTM, thanks!

testbench/database.py
coryan: In the old repo we have this:

https://github.com/googleapis/google-cloud-cpp/blob/f9d58d68ed14d00c583cecd6381fcd6e45c9bae1/google/cloud/storage/emulator/gcs/bucket.py#L209

In this repo we have this:

https://github.com/googleapis/storage-testbench/blob/main/gcs/bucket.py#L211

I think it is a bug to initialize metageneration to `0`, it should be `1`.  That is probably why you see "no" metageneration.

In the old repo we also have this:


https://github.com/googleapis/google-cloud-cpp/blob/f9d58d68ed14d00c583cecd6381fcd6e45c9bae1/google/cloud/storage/emulator/gcs/bucket.py#L273-L274

that is also a bug, the metageneration should be incremented in all changes, regardless of the versioning status.  I changed that already:

https://github.com/googleapis/storage-testbench/blob/85e9d6d39d453ac007455e27aafc58045956293e/gcs/bucket.py#L275

In short.  I think `insert_bucket` *should*, but probably does not, create buckets with a valid metageneration, but that is not a problem in the `database.py` file.

cojenco: I see. Thanks for the clear explanation and fix.

cojenco (approved):

Issue #27, opened 2021-08-05
Cleanups after moving all the code


coryan:
I am noticing potential bugs or code cleanups that should be deferred until I have moved all the code from `google-cloud-cpp`. I will leave TODO entries in the code linking to this bug.

- [ ] Rename emulator.py to testbench.py, may require renaming  `testbench/` too.
- [x] Split `test_emulator.py` it just tests the bucket RPCs and some simple "root" RPCS.
- [ ] #90
- [ ] inserting an object changes the bucket metadata ?
- [ ] the retry instructions call `.popleft()` when maybe they should return `None`?
- [x] Retry control operations return `text/html`, should be `application/json`.
- [ ] The function supporting `PUT` for a resumable upload is too long, it needs refactoring.
- [ ] Maybe the database.supported_methods should be a global
- [ ] "retry tests" is really a generic fault injection mechanism, the name is wrong.

coryan:
I am going to declare this obsolete.  Obviously we are able to maintain the code where it is (as we are doing it).

PR #33, opened 2021-08-13
feat: add support for retry instructions in database

coryan:
Part of the work for #24

### DIFF
--- a/testbench/database.py
+++ b/testbench/database.py
@@ -12,9 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
 import json
 import os
+import re
+import uuid
 
+from schema import Schema, SchemaError, And, Use
 from google.cloud.storage_v1.proto import storage_pb2 as storage_pb2
 
 import gcs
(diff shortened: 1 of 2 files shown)


testbench/database.py
```diff
@@ -256,3 +260,93 @@ def insert_rewrite(self, rewrite):
...
+        ]["instructions"].get(method, None):
+            return self.retry_tests[retry_test_id]["instructions"][method][0]
+        else:
+            return None
+
+    def dequeue_next_instruction(self, retry_test_id, method):
+        self.get_retry_test(retry_test_id)
+        next_instruction = self.retry_tests[retry_test_id]["instructions"][
+            method
+        ].popleft()
```
tritone, line 337: I guess this will return an indexError if the queue is empty, which might translate into a 5xx for the flask response? Do we need to catch this somewhere? Not sure exactly how this function is being used.

tritone (commented):
Nice tests!

testbench/database.py
coryan: > Not sure exactly how this function is being used.

I am not sure either. I am learning about the retry support as I move the classes and implement tests for them :-)

As the code stands today, they check if there are instructions just before using them:

https://github.com/googleapis/google-cloud-cpp/blob/a6329c16e474fe19f0d873d67eed408b26785cc5/google/cloud/storage/emulator/utils/common.py#L382

I might have chosen to return `None` as an indication of "no more instructions", but that sort of cleanup should happen after we move the code.

coryan: Updated #27 to track this.

tritone: Sounds good!

tritone (approved):

PR #37, opened 2021-08-14
feat: add retry wrapper to emulator

coryan:
The emulator wraps each RPC with a "retry test", a mechanism to inject
failures and exercise the retry path in the client libraries.  This
change adds the wrapper to the existing RPCs.

Part of the work for #24

### DIFF
--- a/emulator.py
+++ b/emulator.py
@@ -14,6 +14,7 @@
 
 import flask
 import httpbin
+import json
 from functools import wraps
 from werkzeug.middleware.dispatcher import DispatcherMiddleware
 
@@ -25,7 +26,6 @@ import testbench
 
 db = testbench.database.Database.init()
 grpc_port = 0
-supported_methods = []
 
 # === DEFAULT ENTRY FOR REST SERVER === #
 root = flask.Flask(__name__)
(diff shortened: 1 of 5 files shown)


emulator.py
```diff
@@ -38,6 +38,54 @@ def index():
...
+    response = json.dumps({"retry_test": db.list_retry_tests()})
+    return flask.Response(response, status=200, content_type="application/json")
+
+
+@root.route("/retry_test", methods=["POST"])
+def create_retry_test():
+    payload = json.loads(flask.request.data)
+    test_instruction_set = payload.get("instructions", None)
+    if not test_instruction_set:
+        return flask.Response("instructions is not defined", status=400)
```
cojenco, line 69: Add `content_type="text/plain"` to this response?

testbench/common.py
```diff
@@ -338,6 +341,30 @@ def enforce_patch_override(request):
         testbench.error.notallowed(context=None)
 
 
+def __get_default_response_fn(data):
+    return data
+
+
+def handle_retry_test_instruction(database, request, method):
```
cojenco, line 348: Are we splitting this function and checking in the `error_code_matches` part first?

coryan: I may be misinterpreting your question.  The function looks a bit awkward right now because it just handles non-streaming errors.  It will get bigger eventually, but I am being lazy and not adding this stuff until I strictly need it:

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/storage/emulator/utils/common.py#L380-L424

emulator.py
coryan: Done.

testbench/common.py
cojenco: Yep that was my question; I was just following up on handling retry instructions for connection errors and streaming errors. Your plan SGTM 👍

cojenco (approved):
Nice tests! Thanks for adding them!

PR #42, opened 2021-08-16
feat: emulator add object CRUD operations

coryan:
Part of the work for #24

### DIFF
--- a/tests/test_emulator_object_metadata.py
+++ b/tests/test_emulator_object_metadata.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit test for Object Metadata operations in emulator.py."""
+
+import json
+import os
+import unittest
+from werkzeug.test import create_environ
+
+import emulator
+import gcs
+import testbench
+
+
+class TestEmulatorObjectMetadata(unittest.TestCase):
+    def setUp(self):
+        emulator.db = testbench.database.Database.init()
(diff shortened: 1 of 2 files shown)


tests/test_emulator_object_metadata.py
```diff
@@ -0,0 +1,126 @@
...
+        os.environ.pop("GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME", None)
+
+    def test_object_crud(self):
+        response = self.client.post(
+            "/storage/v1/b", data=json.dumps({"name": "bucket-name"})
+        )
+        self.assertEqual(response.status_code, 200)
+
+        payload = "The quick brown fox jumps over the lazy dog"
+        response = self.client.put(
```
cojenco, line 43: I'm probably not reading this correctly. How/when is the object `fox.txt` created? Why wouldn't this put request return a `notfound` error from [here](https://github.com/googleapis/storage-testbench/blob/6344fee892755508e58feda524b2496ae0195b38/testbench/database.py#L204)?

tests/test_emulator_object_metadata.py
```diff
@@ -0,0 +1,126 @@
...
+            "/storage/v1/b/bucket-name/o/fox.txt",
+            data=json.dumps({"metadata": metadata}),
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(
+            response.headers.get("content-type").startswith("application/json")
+        )
+        patch_rest = json.loads(response.data)
+        self.assertEqual(
+            patch_rest.get("metadata"), patch_rest.get("metadata") | metadata
```
cojenco, line 86: Could you help explain what this assertion does? 🙏

coryan: Apparently `.assertDictContainsSubset(subset, full)` is deprecated (I get a warning).  Instead of do `full == full | subset` (or `subset | full`).  If `subset` is really a subset, then `|` adds nothing and the `==` succeeds :shrug:

tests/test_emulator_object_metadata.py
coryan: It is created right in this line.  It is using the XML `PUT` API.  Too weird to use XML and JSON in the same test?

tests/test_emulator_object_metadata.py
cojenco: Clever and thanks for the explanation!

tests/test_emulator_object_metadata.py
cojenco: Ah I see! Wow I did not realize this is using the XML API 😂

cojenco (approved):
LGTM

PR #47, opened 2021-08-17
feat: add RPCs for Bucket notifications

coryan:
Part of the work for #24

### DIFF
--- a/gcs/bucket.py
+++ b/gcs/bucket.py
@@ -486,6 +486,8 @@ class Bucket:
         return notification
 
     def get_notification(self, notification_id, context):
+        if notification_id not in self.notifications:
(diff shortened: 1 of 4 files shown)


gcs/bucket.py
```diff
@@ -486,6 +486,8 @@ def insert_notification(self, request, context):
         return notification
 
     def get_notification(self, notification_id, context):
+        if notification_id not in self.notifications:
+            testbench.error.notfound("Notification %s" % notification_id, context)
```
cojenco, line 490: Should we add this same check to `delete_notification()` below?

coryan: Done. PTAL.

cojenco (approved):

PR #55, opened 2021-08-19
feat: complete fault injection features

### LINKED ISSUE #54: Storage: Introduce failures into uploads with return-XXX-after-YYYk when using Retry Test API
**What component of `google-cloud-cpp` is this feature request for?**
`google/cloud/storage`

Introduce failures into uploads with return-XXX-after-YYYk when using Retry Test API. It should behave similarly to emulators existing support but should be flexible for values of XXX and YYYk. Where XXX is a HTTP status code and YYYk is a multiple of 256kib

It was marked as a todo since: https://github.com/googleapis/google-cloud-cpp/pull/6611

coryan:
Part of the work for #24. Also fixes #54

### DIFF
--- a/testbench/database.py
+++ b/testbench/database.py
@@ -15,10 +15,8 @@
 import collections
 import json
 import os
-import re
 import uuid
 
-from schema import Schema, SchemaError, And, Use
 from google.cloud.storage_v1.proto import storage_pb2 as storage_pb2
 
 import gcs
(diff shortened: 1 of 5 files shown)


testbench/database.py
```diff
@@ -295,26 +293,24 @@ def get_retry_test(self, retry_test_id):
             testbench.error.notfound("Retry Test %s" % retry_test_id, context=None)
         return self.__to_serializeable_retry_test(retry_test)
 
+    def __validate_injected_failure_description(self, failure):
```
cojenco, line 296: 👍 Looks good, thanks!

tests/test_emulator_retry.py
```diff
@@ -99,6 +102,184 @@ def test_retry_test_return_error(self):
         )
         self.assertEqual(list_response.status_code, 429, msg=list_response.data)
 
+    def test_retry_test_return_error_after(self):
+        response = self.client.post(
+            "/retry_test",
+            data=json.dumps(
+                {"instructions": {"storage.buckets.list": ["return-429-after-128K"]}}
```
cojenco, line 109: I'm still confused. In `handle_retry_test_instruction()` on Line 416, doesn't it only handle `error_after_bytes_matches` when method == "storage.objects.insert"?

cojenco: Ah I see the other `test_retry_test_return_error_after_bytes()` down there 👍

coryan: Do you need me to change something or are we good to go here?

cojenco (approved):

PR #60, opened 2021-08-20
fix: method names in retry_test wrapper

coryan:
I found a couple of methods with the wrong names: one a clear typo, the
onther (`Permission` vs. `Permission*s*`) may be argued both ways. The
plural form matches the name in:

https://cloud.google.com/storage/docs/json_api/v1/buckets/testIamPermissions

So I went with that.

### DIFF
--- a/tests/test_emulator_retry.py
+++ b/tests/test_emulator_retry.py
@@ -36,6 +36,63 @@ class TestEmulatorRetry(unittest.TestCase):
         response = self.client.get("/")
         self.assertEqual(response.data, b"OK")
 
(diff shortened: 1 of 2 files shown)


tests/test_emulator_retry.py
```diff
@@ -36,6 +36,63 @@ def test_root(self):
         response = self.client.get("/")
         self.assertEqual(response.data, b"OK")
 
+    def test_retry_test_supported_operations(self):
```
cojenco, line 39: Great test! Thanks for adding this validation!

cojenco (approved):
LGTM

PR #63, opened 2021-08-20
feat: add raise_error route

coryan:
I am not sure where this is used (not in the C++ client library, and in
none of the existing tests), but it seems plausible that somehow it is
used in some other client library tests.

Part of the work for #24

### DIFF
--- a/tests/test_emulator_root.py
+++ b/tests/test_emulator_root.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


tests/test_emulator_root.py
```diff
@@ -0,0 +1,50 @@
...
+        emulator.db.clear()
+        self.client = emulator.server.test_client()
+        # Avoid magic buckets in the test
+        os.environ.pop("GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME", None)
+
+    def test_root(self):
+        response = self.client.get("/")
+        self.assertEqual(response.data, b"OK")
+
+    def test_raise_defaul(self):
```
cojenco, line 36:
```suggestion
    def test_raise_default(self):
```

tests/test_emulator_root.py
```diff
@@ -0,0 +1,50 @@
...
+        response = self.client.get("/raise_error", query_string={"msg": "test-only"})
```
cojenco, line 37: Q: Would we be able to test this with `with self.assertRaises(...) as exc:`

coryan: You mean calling the `emulator.raise_error()` function directly?  I thought the interesting bit was testing that the route returned the appropriate error.  Happy to also test that if you want.

cojenco: Just curious. This is sufficient and good to go!

cojenco (approved):

PR #65, opened 2021-08-20
chore: support Python < 3.9

### LINKED ISSUE #64: Create a CI build with Python 3.*
We need to test with at least Python 3.8, maybe others.  Some of the tests may be using Python 3.9 features.

coryan:
Python < 3.9 lacks dict union operators:

https://www.python.org/dev/peps/pep-0584

All the alternatives are ugly (otherwise they would not have added the
operator!). I decided to use the `{**a, **b}` formulation.

Fixes #64

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


cojenco (approved):
Thanks for checking and patching 😄

Issue #58, opened 2021-08-20
Move gRPC implementation to v2/ protos


coryan:
The current implementation is based on the storage/v1 protos, we need to move the testbench to storage/v2.

PR #71, opened 2021-08-21
chore: introduce storage/v2 generated files

coryan:
We want to migrate the code to use the storage/v2 protos. I want to do
this in relatively small PRs, which will probably require having both
the storage/v1 and storage/v2 files available for a bit.

I do not think we need to create a special branch in the
`googleapis/python-storage` repository, this repository is a perfectly
suitable location for the protos (at least for now).

I also documented how to regenerate these files in the future.

Part of the work for #58


cojenco (approved):
Thanks for documenting how to regenerate the proto files too 🙏

PR #73, opened 2021-08-24
feat: add project related endpoints to emualator retry tests

### LINKED ISSUE #56: Storage(emulator): Add project related endpoints to Retry Test API
**What component of `google-cloud-cpp` is this feature request for?**
GCS `google/cloud/storage/emulator`

Per discussion with @frankyn, incorporate project related routes in [project.py](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/storage/emulator/gcs/project.py) into Retry Test API. 

The [Retry Test API](https://github.com/googleapis/google-cloud-cpp/tree/main/google/cloud/storage/emulator#retry-test-api) offers a mechanism to describe more complex retry scenarios while sending a single, constant header through all the HTTP requests from a test program. It is used to enable the Retry Strategy Conformance Tests. Currently, the Retry Test API supports bucket and object related operations. The Retry Test API should also cover project-related operations in order to verify if retries are working as intended across all operations.

Examples of project related operations include [`Projects.hmacKeys`](https://cloud.google.com/storage/docs/json_api/v1/projects/hmacKeys) and [`Projects.serviceAccount`](https://cloud.google.com/storage/docs/json_api/v1/projects/serviceAccount)

Note: This issue is planned to be addressed in the new storage-testbench repo

Breathtender:
Fixes #56

### DIFF
--- a/emulator.py
+++ b/emulator.py
@@ -886,6 +886,61 @@ def delete_resumable_upload(bucket_name):
     db.delete_upload(upload_id, None)
     return flask.make_response("", 499, {"content-length": 0})
 
+# PROJECT ROUTES #
+
+@gcs.route("/projects/<project_id>/hmacKeys", methods=["POST"])
+@retry_test("storage.hmacKey.create")
+def hmackeys_create(project_id):
+    project = gcs_type.project.get_project(project_id)
+    email = flask.request.args.get("email")
+    response = project.insert_hmac_key(email)
+    fields = flask.request.args.get("fields", None)
+    return testbench.common.filter_response_rest(response, None, fields)
+
+@gcs.route("/projects/<project_id>/hmacKeys/<access_id>", methods=["DELETE"])
+@retry_test("storage.hmacKey.delete")
+def hmackeys_delete(project_id, access_id):
+    project = gcs_type.project.get_project(project_id)
+    project.delete_hmac_key(access_id)
+    return ""
+
+@gcs.route("/projects/<project_id>/hmacKeys/<access_id>", methods=["GET"])
+@retry_test("storage.hmacKey.get")
+def hmackeys_get(project_id, access_id):
+    project = gcs_type.project.get_project(project_id)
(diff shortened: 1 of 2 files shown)


emulator.py
```diff
@@ -886,6 +886,61 @@ def delete_resumable_upload(bucket_name):
     db.delete_upload(upload_id, None)
     return flask.make_response("", 499, {"content-length": 0})
 
+# PROJECT ROUTES #
+
+@gcs.route("/projects/<project_id>/hmacKeys", methods=["POST"])
```
coryan, line 891: Seems like this duplicates the code in:

https://github.com/googleapis/storage-testbench/blob/dfaed891836b1ef8152c79133dc3f5187ec8c954/gcs/project.py#L194-L287

We can either refactor the  `retry_test()` wrapper so it is usable from there, or we can move the functions from that location to here.

gcs/project.py
```diff
@@ -182,6 +182,10 @@ def get_hmac_key(self, access_id):
             testbench.error.notfound("service account for key=%s" % access_id, None)
         return sa.get_key(key_id)
 
+    def list_hmac_keys(self):
+        """List keys across all service accounts in the project"""
+        return [sa.key_items() for sa in self.service_accounts.items() for sa in sa]
```
coryan, line 187: Please add a test to make sure this is covered, it is a relative simple function, but let's try to keep the coverage high (it avoids hiding other errors).

emulator.py
coryan: BTW, if you decide to keep the functions in `projects.py`, then the existing unit tests will work.  Otherwise, please remember to change those unit tests to target the new functions.

coryan (changes requested):
Thanks for this change!  There is some duplication with existing code, more details in the comments.

emulator.py
Breathtender: Oh sweet that makes sense, will work to integrate those then.

PR #91, opened 2021-08-30
refactor: move entry points to testbench module

coryan:
This is part of the effort to remove any references to "emulator" in the
repository. I left `emulator.py` around for backwards compatibility, but
I suspect we can remove it after checking with a few folks.

Part of the work for #27 

The line count looks scary: I renamed `emulator.py` to `testbench/rest_server.py`, and then added some small functions for backwards compatibility.


testbench/__main__.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/python3
+#
+# Copyright 2020 Google LLC
```
cojenco, line 3: Thanks for the thorough scrub and move! 
Just one nit and q: Did we ever decide on keeping the original date or updating to 2021 for changes like these?

testbench/rest_server.py
```diff
@@ -0,0 +1,897 @@
+# Copyright 2020 Google LLC
```
cojenco, line 1: Same question here

coryan: Fixed.  I thought `git` was going to treat that as a move, but I never argue with the robots. Fixed, and reverted the change in `emulator.py`.

testbench/__main__.py
coryan: This is a mistake, fixed.

coryan:
Rebased, resolved conflicts and addressed your comments. PTAL.

cojenco (approved):
LGTM👍 Thanks Carlos!

PR #92, opened 2021-08-30
feat: complete gRPC server implementation

coryan:
Part of the changes for #58, the only thing missing is a REST API to start the gRPC server.

### DIFF
--- a/tests/test_grpc_server.py
+++ b/tests/test_grpc_server.py
@@ -168,15 +168,23 @@ class TestGrpc(unittest.TestCase):
 
         def streamer():
             media = TestGrpc._create_block(517 * 1024).encode("utf-8")
-            id = start.upload_id
-            step = storage_pb2.ServiceConstants.Values.MAX_READ_CHUNK_BYTES
-            for offset in range(0, len(media), step):
-                upload_id = id
-                id = None
+            step = 256 * 1024
+            end = min(step, len(media))
+            content = media[0:end]
+            # The first message is special, it should have a an upload id.
+            yield storage_pb2.WriteObjectRequest(
+                upload_id=start.upload_id,
+                write_offset=0,
+                checksummed_data=storage_pb2.ChecksummedData(
+                    content=content, crc32c=crc32c.crc32c(content)
+                ),
+                finish_write=(end == len(media)),
+            )
+
+            for offset in range(step, len(media), step):
                 end = min(offset + step, len(media))
                 content = media[offset:end]
                 yield storage_pb2.WriteObjectRequest(
-                    upload_id=upload_id,
                     write_offset=offset,
                     checksummed_data=storage_pb2.ChecksummedData(
                         content=content, crc32c=crc32c.crc32c(content)
@@ -239,7 +247,7 @@ class TestGrpc(unittest.TestCase):
         )
 
         offset = 2 * QUANTUM
-        content = media[2 * QUANTUM:]
+        content = media[2 * QUANTUM :]
         r3 = storage_pb2.WriteObjectRequest(
             write_offset=QUANTUM,
             checksummed_data=storage_pb2.ChecksummedData(
--- a/tests/test_grpc_server.py
+++ b/tests/test_grpc_server.py
@@ -19,7 +19,9 @@
 import crc32c
 import json
 import unittest
-from google.storage.v2 import storage_pb2
+import unittest.mock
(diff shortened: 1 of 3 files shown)


tests/test_grpc_server.py
```diff
@@ -71,6 +73,263 @@ def test_read_object(self):
...
+        )
+        self.assertIsNotNone(start.upload_id)
+
+        def streamer():
+            media = TestGrpc._create_block(517 * 1024).encode("utf-8")
+            id = start.upload_id
+            step = storage_pb2.ServiceConstants.Values.MAX_READ_CHUNK_BYTES
+            for offset in range(0, len(media), step):
+                upload_id = id
+                id = None
```
cojenco, line 175: What is the use for `id`?

tests/test_grpc_server.py
```diff
@@ -71,6 +73,263 @@ def test_read_object(self):
...
+            upload_id=start.upload_id,
+            write_offset=offset,
+            checksummed_data=storage_pb2.ChecksummedData(
+                content=content, crc32c=crc32c.crc32c(content)
+            ),
+            finish_write=False,
+        )
+
+        offset = 2 * QUANTUM
+        content = media[QUANTUM:]
```
cojenco, line 242:
```suggestion
        content = media[2 * QUANTUM:]
```

tests/test_grpc_server.py
coryan: Only the first `WriteObjectRequest` can have a `upload_id` field set.  So, after the first iteration `id` is `None`, and therefore only on the first iteration `upload_id` has a valid value.  Terrible?  Maybe.

Is there a more elegant way to yield one value that is "special" and then yield other values?  Oh... can I yield before entering the loop?

tests/test_grpc_server.py
coryan: Thanks!

tests/test_grpc_server.py
cojenco: Not sure I follow... why is it that we only want the first `WriteObjectRequest` to have a valid `upload_id`?

coryan: I think the proto spec requires that:

https://github.com/googleapis/googleapis/blob/68e5bf3045d7b3ef55f473494a0cd4566491f899/google/storage/v2/storage.proto#L222-L232

we may not be enforcing it though.

coryan: I change the code to yield once before entering the loop and added a comment.

tests/test_grpc_server.py
```diff
@@ -71,6 +73,271 @@ def test_read_object(self):
...
+                content = media[offset:end]
+                yield storage_pb2.WriteObjectRequest(
+                    write_offset=offset,
+                    checksummed_data=storage_pb2.ChecksummedData(
+                        content=content, crc32c=crc32c.crc32c(content)
+                    ),
+                    finish_write=(end == len(media)),
+                )
+
+        write = self.grpc.WriteObject(streamer(), "fake-context")
```
cojenco, line 195: I haven't user generators much, but I think we need a for loop in order to generate (and use) each `WriteObjectRequest`

IIUC, `streamer()` is just a generator object. To yield each `WriteObjectRequest`, we either need to utilize a for loop or call `next()` on the generator object.

tests/test_grpc_server.py
cojenco: I see! Thanks for the change. Another question on line 195 :)

tests/test_grpc_server.py
cojenco: Oh wait does `WriteObject` consume a stream-like generator?

coryan: So it works right?  I just tested with:

```py
def foo():
...     yield 1
...     yield 2
...     for i in range(10, 15): yield i
... 
>>> 
>>> 
>>> g = foo()
>>> l = list(g)
>>> print(l)
[1, 2, 10, 11, 12, 13, 14]
```

That is what I expected.

cojenco: Yep looks good! Sorry about the confusion 😂

cojenco (approved):
LGTM, thanks!

tests/test_grpc_server.py
coryan: Thanks to you.

PR #97, opened 2021-08-31
feat: add support files for packaging

coryan:
The intent is to use this with a github URL from `google-cloud-cpp`, or
from any other repository with similar needs. We do *NOT* anticipate
publishing this to PyPI.

This is my first time creating a Python package. It is likely that I am doing
something wrong.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,1 @@
-git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
-grpcio==1.38.1
-grpcio-tools==1.38.1
-googleapis-common-protos==1.53.0
-protobuf==3.17.3
-flask==2.0.1
-httpbin==0.7.0
-requests-toolbelt==0.9.1
-scalpl==0.4.2
-crc32c==2.2
-gunicorn==20.1.0
-schema==0.7.4
+.
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,53 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -0,0 +1,53 @@
...
+    url="https://github.com/googleapis/storage-testbench",
+    project_urls={
+        "Bug Tracker": "https://github.com/googleapis/storage-testbench/issues",
+    },
+    classifiers=[
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: Apache-2.0",
+        "Operating System :: OS Independent",
+    ],
+    package_dir={"": "."},
```
cojenco, line 37: Perhaps we don't need to declare the optional `package_dir` since we're following a common convention?

coryan: SGTM.  Thanks.


Any thoughts on the package name (`googleapis-storage-testbench`) and the fact that it creates a `testbench` module?

It also creates  a `gcs` and `google/storage/v2` modules, but we will remove those eventually.

requirements.txt
```diff
@@ -1,12 +1 @@
...
-googleapis-common-protos==1.53.0
-protobuf==3.17.3
-flask==2.0.1
-httpbin==0.7.0
-requests-toolbelt==0.9.1
-scalpl==0.4.2
-crc32c==2.2
-gunicorn==20.1.0
-schema==0.7.4
+.
```
cojenco, line 1: Thanks for adding the support files! Will we be removing this `requirements.txt` file entirely once finalized? (Meaning that all users should install the testbench with `pip`?)  If so, we'll also need to update the readme instructions.

setup.py
cojenco: ^^ I don't think that would be a problem. I've been looking, comparing/contrasting with other googleapis-python repositories, and it's done similarly 😬

Once we change the structure, `/testbench` will kind of be the overarching directory, then we might want to set `package_dir`. Though it doesn't seem necessary since it's relatively easy to declare the packages.

requirements.txt
coryan: I honestly do not know.  We could change the instructions in this PR, I would be Okay with that.  I am Okay with pushing to a future PR.

Right now, `pip install -e .` works: it installs the dependencies needed for development.  But arguably `pip install -r requirements.txt` installs too much.

cojenco: That's good to know, thanks! I'll have to look into pip more. I agree we can move this forward and push instruction changes to a future PR.

cojenco (approved):
Thanks for adding pip packaging support!

PR #98, opened 2021-09-01
cleanup: sort import directives

coryan:
The recommended order seems to be standard library modules, then
packages, then local modules.

### DIFF
--- a/gcs/holder.py
+++ b/gcs/holder.py
@@ -14,16 +14,15 @@
 
 """Implement a holder for resumable upload's data and rewrite's data"""
 
-from typing import ClassVar
-import crc32c
 import flask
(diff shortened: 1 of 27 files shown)


gcs/holder.py
```diff
@@ -14,16 +14,15 @@
 
 """Implement a holder for resumable upload's data and rewrite's data"""
 
-from typing import ClassVar
-import crc32c
 import flask
```
cojenco, line 17: Thanks for the cleanup! `flask` should be moved down?

coryan: Fixed.

cojenco (approved):
LGTM, thanks!!

PR #106, opened 2021-09-01
fix: use locks when accessing the database

coryan:
The database object is a global, shared by many threads. While we could
get away without synchronization (because there is no I/O in it to cause
a thread switch), we shouldn't.

Also renamed all the database fields to ensure they are always accessed
via a method that has the right locking. This is where near 100% code
coverage comes handy.


cojenco (approved):
LGTM👍 Thanks for adding this!

PR #113, opened 2021-09-02
fix: fields in objects.rewrite response should be strings

### LINKED ISSUE #112: Response from object rewrite incorrect
Current response fields `totalBytesRewritten` and `objectSize` are numbers, should be strings according to the json api: https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite#response

BrennaEpp:
Fixes #112


coryan (approved):
Thanks!

PR #122, opened 2021-09-10
fix: add retention_policy.effective_time when locking a policy

### LINKED ISSUE #116: RetentionPolicy.EffectiveTime not set by server
The testbench does not set the `effective_time` of a RetentionPolicy when there is a call to `buckets.patch` updating the retention policy. The response then doesn't include `effectiveTime: x` which causes a parse error in the Go client as we try to parse the `effectiveTime` field.

A call to `buckets.insert` with a retention policy attached has the same issue: the effective time is not set by the testbench.

Note that a call to `buckets.lockRetentionPolicy` also does not set the `effectiveTime` nor `retentionPeriod` fields which also leads to parse errors. Not sure what the correct behaviour should be for this method if the bucket has no retention policy when attempting to lock it.

BrennaEpp:
Fixes #116 for the Go conformance tests, which use the lock method and need an effective time to parse.

### DIFF
--- a/testbench/rest_server.py
+++ b/testbench/rest_server.py
@@ -426,7 +426,9 @@ def bucket_test_iam_permissions(bucket_name):
 def bucket_lock_retention_policy(bucket_name):
     bucket = db.get_bucket(flask.request, bucket_name, None)
     bucket.metadata.retention_policy.is_locked = True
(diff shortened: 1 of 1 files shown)


tritone (approved):
Looks good once you fix the linting.

PR #125, opened 2021-09-21
fix: cleanup filter_response_rest and nested_keys

Breathtender:
A bug was happening where fields wasn't working properly and it was
leaving nested empty dicts in the response. This should clear that up,
by changing nested_keys function so that it includes parent keys also.


coryan (approved):

cojenco (approved):
LGTM, thanks!

PR #129, opened 2021-09-24
fix: missing directory is setup.py

coryan:
The pip package has been broken since #114, I just did not notice because my builds were pinned to v0.4.0.


Breathtender (approved):
nice catch

PR #172, opened 2021-12-01
cleanup: remove unused dependency

coryan:


frankyn (approved):
Ah, did I add it and never use it? I thought I did some validation with this package, but there is no code. Must have been an accident then. Thanks for catching it!

coryan:
You used it, I removed the usage in favor of "simpler" code.

PR #175, opened 2021-12-04
feat: support additional x-goog-* headers

coryan:
The testbench was not generating some `x-goog-*` headers. I will need
these headers in the C++ client library.


coryan:
ping

tritone (approved):

PR #179, opened 2021-12-13
chore: update protos circa 2021-12-13

coryan:
Use the preview branch for the protos, as it contains the RPO and CDR
changes, as well as many RPCs that we need to start working on.


coryan:
I am going to ignore the drop in coverage.  We will fix that as we implement more RPCs in `grpc_server.py`.

tritone (approved):

PR #211, opened 2021-12-16
refactor: rename DataHolder to Upload

### LINKED ISSUE #22: Split gcs.holder.DataHolder into two classes
I am about to send a PR adding the `gcs.holder.DataHolder` class, which is used both for resumable uploads and for rewrites. We should have separate classes for each.

coryan:
Fixes #22

### DIFF
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Implement a holder for resumable upload's data and rewrite's data"""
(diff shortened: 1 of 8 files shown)


gcs/upload.py
```diff
@@ -27,14 +27,21 @@
 import testbench
 
 
-class DataHolder(types.SimpleNamespace):
+class Upload(types.SimpleNamespace):
```
frankyn, line 30: Internal name is dropzone or something like that; this is fine.

frankyn (approved):

PR #212, opened 2021-12-16
feat: support gRPC for Rewrite objects

coryan:
We need to support gRPC-created Rewrite objects.

Part of the work for #192

### DIFF
--- a/gcs/rewrite.py
+++ b/gcs/rewrite.py
@@ -14,16 +14,9 @@
 
 """Helper class for rewrite operations."""
 
-import hashlib
-import json
 import types
 import uuid
 
-import crc32c
-import flask
-from google.protobuf import json_format
-
-from google.storage.v2 import storage_pb2
 import testbench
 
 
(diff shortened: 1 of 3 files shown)


gcs/rewrite.py
```diff
@@ -56,25 +60,64 @@ def init_rest(
...
-        token = hashlib.sha256(
-            (
-                "%s/o/%s/rewriteTo/b/%s/o/%s"
-                % (src_bucket_name, src_object_name, dst_bucket_name, dst_object_name)
-            ).encode("utf-8")
-        ).hexdigest()
+
+    @classmethod
+    def init_grpc(cls, request, context):
+        if not request.source_bucket.startswith("projects/_/buckets/"):
```
tritone, line 78: Does JSON have this validation logic? If not should it be shared?

tritone (commented):
Generally looks good, one question.

gcs/rewrite.py
coryan: JSON bucket names are not prefixed by `projects/_/buckets/` so the validation cannot be identical. I assume you are asking about the other checks (non-empty and so forth).  I think the bucket names being non-empty is validated implicitly by the Flask route decorators, e.g.:

https://github.com/googleapis/storage-testbench/blob/c1d3ec4e96a519e1b1f1d6077eaf96575678fb5c/testbench/rest_server.py#L595-L600

tritone: Yep that's correct-- sounds good, thanks!

tritone (approved):

Issue #217, opened 2021-12-21
Simplify handling of preconditions and generations


coryan:
Currently the `testbench.database.Database` class tries to handle all requests "generically", and extracts the bits of data that it needs from them using helper functions.

These functions are fairly hard to reason about, they do things like "if the context is None, then it is probably a REST request, and then probably the generation is in the `generation` argument".  And in other places we convert XML requests into a `FakeRequest` just so we can normalize the pre-conditions from headers to query parameters.

I think it makes more sense for the caller into the `Database.*` methods to pass the generation and/or preconditions.  They know where the right attributes are after all.

- [x] Helper functions to create precondition predicates for object requests
- [x] Helper functions to create precondition predicates for bucket requests
- [x] `Database.insert_object()`
- [x] `Database.get_object()`
- [x] `Database.delete_object()`
- [x] `Database.insert_object()`
- [x] `Database.get_bucket*()`
- [x] `Database.delete_bucket()`

PR #218, opened 2021-12-21
cleanup: simplify pre-condition handling

coryan:
This simplifies pre-condition handling for `Database.get_object()` and
`Database.delete_object()` by asking the caller to provide the
generation (which can be `None` or `0`) and the preconditions as a list
of predicates on the blob and the live generation.

The callers know what the pre-conditions fields are called (e.g.
`ifGenerationMatch` vs. `ifSourceGenerationMatch` for `Objects: copy` or
`Objects: rewrite`). The caller also knows if the generation is a
top-level attribute of the request, or part of a sub-object (or a query
parameter or some other possibilities).

Part of the work for #217

### DIFF
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -311,6 +311,105 @@ def extract_media(request):
     return request.data
 
 
+def make_json_preconditions(request, prefix="if"):
+    """Create the pre-conditions for most JSON requests.
+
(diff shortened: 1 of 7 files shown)


coryan:
Sorry for the large PR. The tests are a bit verbose, and I was more interested in clarity than being super brief.

coryan:
I think I need to break down this PR differently.

PR #219, opened 2021-12-21
feat: helper functions for preconditions

coryan:
This is the first step to refactor precondition handling and simplify
how the testbench database is implemented.

Part of the work for #217

### DIFF
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -311,6 +311,205 @@ def extract_media(request):
     return request.data
 
 
+def make_json_preconditions(request, prefix="if"):
+    """Create the pre-conditions for most JSON requests.
+
+    The prefix parameter can be `if` or `ifSource` to handle the special names
+    used for source pre-conditions in `Objects: copy` and `Objects: rewrite`.
+    """
+
+    def if_metageneration_match(blob, _, ctx):
+        match = request.args.get(prefix + "MetagenerationMatch", None)
+        assert match is not None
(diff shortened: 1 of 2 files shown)


testbench/common.py
```diff
@@ -311,6 +311,205 @@ def extract_media(request):
...
+def make_json_preconditions(request, prefix="if"):
+    """Create the pre-conditions for most JSON requests.
+
+    The prefix parameter can be `if` or `ifSource` to handle the special names
+    used for source pre-conditions in `Objects: copy` and `Objects: rewrite`.
+    """
+
+    def if_metageneration_match(blob, _, ctx):
+        match = request.args.get(prefix + "MetagenerationMatch", None)
+        assert match is not None
```
tritone, line 323: Why have this assertion separately? I think if you skip the default `None` in the line above it will just throw a KeyError there.

tritone (approved):
One question, otherwise LGTM

testbench/common.py
coryan: I must admit I was just cut&pasting previous code I have written here, but I think this is correct. `flask.Request` returns `None` by default:

https://flask.palletsprojects.com/en/2.0.x/api/#flask.Request.args

https://werkzeug.palletsprojects.com/en/2.0.x/datastructures/#werkzeug.datastructures.MultiDict.get

By construction the `None` value is not possible, an assertion is mostly to make the code easier to read / follow.

PR #223, opened 2021-12-28
cleanup: simplify bucket precondition handling

### LINKED ISSUE #217: Simplify handling of preconditions and generations

coryan:
Fixes #217


coryan:
I think we can live with the reduced code coverage as it comes from also reducing the amount of code.

tritone (approved):
nice cleanup!

PR #228, opened 2021-12-29
feat(grpc): implement GetIamPolicy()

### LINKED ISSUE #194: `rpc GetIamPolicy(google.iam.v1.GetIamPolicyRequest) returns (google.iam.v1.Policy)`

coryan:
Fixes #194

### DIFF
--- a/testbench/grpc_server.py
+++ b/testbench/grpc_server.py
@@ -47,6 +47,11 @@ class StorageServicer(storage_pb2_grpc.StorageServicer):
         self.db.insert_bucket(bucket, context)
         return bucket.metadata
 
+    def GetIamPolicy(self, request, context):
+        self.db.insert_test_bucket()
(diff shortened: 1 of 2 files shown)


testbench/grpc_server.py
```diff
@@ -47,6 +47,11 @@ def CreateBucket(self, request, context):
         self.db.insert_bucket(bucket, context)
         return bucket.metadata
 
+    def GetIamPolicy(self, request, context):
+        self.db.insert_test_bucket()
```
tritone, line 51: Not related to this PR really, but could we factor out this call to happen on server startup maybe?

tritone (approved):

testbench/grpc_server.py
coryan: Good idea, will do.

PR #274, opened 2022-02-02
feat: add support for return-X-after-YK of downloaded data

### LINKED ISSUE #273: Support return-X-after-YK for downloads
The Retry Test API currently supports `return-X-after-YK` for uploads. To expand testing scenarios in the retry conformance tests, we want to add support in the Retry Test API to `return-X-after-YK` of downloaded data.

cojenco:
Fixes #273

As part of the GCS client library retry alignment project, this adds support in the Retry Test API to `return-X-after-YK` of **downloaded** data.

I've noted a TODO entry to introduce a Retry Test API method "storage.objects.download" to further support complex download test cases for larger files.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ curl -H "x-retry-test-id: 1d05c20627844214a9ff7cbcf696317d" "http://localhost:91
 | Failure Id              | Description
 | ----------------------- | ---
 | return-X                                  | Testbench will fail with HTTP code provided for `X`, e.g. return-503 returns a 503
-| return-X-after-YK                         | Testbench will return X after YKiB of uploaded data
-| return-broken-stream-final-chunk-after-YB | Testbench will break connection on final chunk of a resumable upload after Y bytes.
+| return-X-after-YK                         | Testbench will return X after YKiB of uploaded or downloaded data
+| return-broken-stream-final-chunk-after-YB | Testbench will break connection on final chunk of a resumable upload after Y bytes
 | return-broken-stream                      | Testbench will fail after a few bytes
 | return-reset-connection                   | Testbench will fail with a reset connection
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -677,17 +677,17 @@ def __get_streamer_response_custom_error_fn(database, method, test_id, error_cod
             d = _extract_data(data)
             bytes_downloaded = 0
             for r in range(0, len(d), chunk_size):
+                if bytes_downloaded < limit:
+                    chunk_end = min(r + chunk_size, len(d))
+                    chunk_downloaded = chunk_end - r
+                    bytes_downloaded += chunk_downloaded
+                    yield d[r:chunk_end]
                 if bytes_downloaded >= limit:
                     database.dequeue_next_instruction(test_id, method)
                     raise testbench.error.RestException(
                         f"Retry Test: Caused a {error_code}. Fault injected after downloading {bytes_downloaded} bytes",
                         error_code
                     )
-                else:
-                    chunk_end = min(r + chunk_size, len(d))
-                    chunk_downloaded = chunk_end - r
-                    bytes_downloaded += chunk_downloaded
-                    yield d[r:chunk_end]
 
         # Inject fault outside of the streamer considering execution flow and werkzeug wrappers
         try:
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -242,7 +242,7 @@ curl -H "x-retry-test-id: 1d05c20627844214a9ff7cbcf696317d" "http://localhost:91
 | Failure Id              | Description
 | ----------------------- | ---
 | return-X                                  | Testbench will fail with HTTP code provided for `X`, e.g. return-503 returns a 503
-| return-X-after-YK                         | Testbench will return X after YKiB of uploaded data
-| return-broken-stream-final-chunk-after-YB | Testbench will break connection on final chunk of a resumable upload after Y bytes.
+| return-X-after-YK                         | Testbench will return X after YKiB of uploaded or downloaded data
```
coryan, line 245: nit: `return-X-after-YK` cannot really work for a download... the error code is returned at the beginning, after the headers but before the payload.  Any client will receive a 200 if the download starts successfully.

testbench/common.py
```diff
@@ -671,6 +671,37 @@ def streamer():
...
+            for r in range(0, len(d), chunk_size):
+                if bytes_downloaded < limit:
+                    chunk_end = min(r + chunk_size, len(d))
+                    chunk_downloaded = chunk_end - r
+                    bytes_downloaded += chunk_downloaded
+                    yield d[r:chunk_end]
+                if bytes_downloaded >= limit:
+                    database.dequeue_next_instruction(test_id, method)
+                    raise testbench.error.RestException(
+                        f"Retry Test: Caused a {error_code}. Fault injected after downloading {bytes_downloaded} bytes",
```
coryan, line 690: For a large buffer, and probably only when using a real TCP/IP connection, the error code here has no effect on the client-side, just FYI.

testbench/common.py
```diff
@@ -752,6 +783,14 @@ def handle_retry_test_instruction(database, request, method):
                     grpc_code=StatusCode.INTERNAL,  # not really used
                     context=None,
                 )
+    # TODO(cathyo@): introduce Retry Test API method "storage.objects.download" for large file downloads
```
coryan, line 786: Please use a GitHub issue numbers to track TODO items, in case you win the lottery and or leave us for greener pa$ture$

tests/test_testbench_retry.py
```diff
@@ -442,6 +442,59 @@ def test_retry_test_return_error_after_bytes(self):
...
 
+    def test_retry_test_return_error_after_bytes_downloaded(self):
+        response = self.client.post(
+            "/storage/v1/b", data=json.dumps({"name": "bucket-name"})
+        )
+        self.assertEqual(response.status_code, 200)
+        # Use the XML API to inject a larger object and smaller object.
+        media = self._create_block(UPLOAD_QUANTUM)
+        blob_larger = self.client.put(
+            "/bucket-name/256kb.txt",
```
coryan, line 453: nit: some inconsistency between the filename with `kb` (Kilo**bits**) and the generated block (`256KiB`).

coryan (changes requested):
A few nits. The `TODO` entry I care about, the stuff about file names and `KiB` vs. `KB` vs. `kb` is just trivia.

I do wonder if this is going to work as you expect outside of unit tests.

README.md
cojenco: I see. I did not think of the fact that the error code is returned before the payload. So `return-X-after-YK` is unrealistic to happen for downloads in the live server...

I think it "works" here merely because I forcefully trigger the streamer first before returning the response (L694)
```
# Inject fault outside of the streamer considering execution flow and werkzeug wrappers
try:
    response = flask.Response(streamer())
    _ = response.data
    return response
except testbench.error.RestException as e:
    raise e
 ```
Thus it "works" when I test it against the python retry conformance test, like so
```
127.0.0.1 - - [02/Feb/2022:16:50:50 -0800] "GET /download/storage/v1/b/aef6bb4577734e59b6f41c5cdfe501fa/o/4678f04722e54979845541c381a5025e?alt=media HTTP/1.1" 504 97 "-" "python-requests/2.27.1"
127.0.0.1 - - [02/Feb/2022:16:50:51 -0800] "GET /download/storage/v1/b/aef6bb4577734e59b6f41c5cdfe501fa/o/4678f04722e54979845541c381a5025e?alt=media HTTP/1.1" 200 1100000 "-" "python-requests/2.27.1"
```

I will probably close this FR since it does not match live server behavior, unless we decide it is helpful in an way for testing. Thanks for the feedback!

PR #276, opened 2022-02-09
feat: add support for return-broken-stream-after-YK

### LINKED ISSUE #173: Support return-broken-stream-after-YB
We currently have `return-broken-stream` which breaks stream after a constant number of bytes that's defined in the emulator. We should have a way to break stream after a certain number of bytes.

cojenco:
Fixes #173

This introduces support for `return-broken-stream-after-YK` of downloaded data in the Retry Test API. 
To better test complex download scenarios in the retry conformance tests, we've discussed to test larger files and add support for "after-YK" instead of "after-YB"

- [X] Tests pass
- [X] Appropriate changes to README are included in PR

### DIFF
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -736,7 +736,7 @@ def handle_retry_test_instruction(database, request, method):
             raise testbench.error.RestException(
                 "Injected 'connection reset by peer' fault", 500
             )
-        elif items[0] == "broken-stream" and method == "storage.objects.get":
+        elif items[0] == "broken-stream":
             conn = request.environ.get("gunicorn.socket", None)
             return __get_streamer_response_fn(database, method, conn, test_id)
     broken_stream_after_bytes = (
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -664,12 +664,12 @@ def __get_streamer_response_fn(database, method, conn, test_id, limit=4, chunk_s
                         conn.close()
                     # This exception is raised to abort the flow control. The
                     # connection has already been closed, causing the client to
-                    # receive a "connection reset by peer" (or a similar error).
+                    # receive a "broken stream" or "connection reset by peer" error.
                     # The exception is useful in unit tests (where there is no
                     # socket to close), and stops the testbench from trying to
                     # complete a request that we intentionally aborted.
                     raise testbench.error.RestException(
-                        "Injected 'connection reset by peer' fault", 500
+                        "Injected 'broken stream' fault", 500
                     )
 
         return flask.Response(streamer())
@@ -746,7 +746,9 @@ def handle_retry_test_instruction(database, request, method):
         items = list(broken_stream_after_bytes.groups())
         after_bytes = int(items[0]) * 1024
         conn = request.environ.get("gunicorn.socket", None)
-        return __get_streamer_response_fn(database, method, conn, test_id, after_bytes)
+        return __get_streamer_response_fn(
+            database, method, conn, test_id, limit=after_bytes
+        )
     error_after_bytes_matches = testbench.common.retry_return_error_after_bytes.match(
         next_instruction
     )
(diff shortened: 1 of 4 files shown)


coryan (approved):

testbench/common.py
```diff
@@ -733,6 +739,14 @@ def handle_retry_test_instruction(database, request, method):
...
             conn = request.environ.get("gunicorn.socket", None)
             return __get_streamer_response_fn(database, method, conn, test_id)
+    broken_stream_after_bytes = (
+        testbench.common.retry_return_broken_stream_after_bytes.match(next_instruction)
+    )
+    if broken_stream_after_bytes and method == "storage.objects.get":
+        items = list(broken_stream_after_bytes.groups())
+        after_bytes = int(items[0]) * 1024
+        conn = request.environ.get("gunicorn.socket", None)
+        return __get_streamer_response_fn(database, method, conn, test_id, after_bytes)
```
tritone, line 749: Nit: Use keyword arg rather than positional for limit.

testbench/common.py
```diff
@@ -644,27 +647,30 @@ def _extract_data(data):
...
+def __get_streamer_response_fn(database, method, conn, test_id, limit=4, chunk_size=4):
     def response_handler(data):
         def streamer():
-            database.dequeue_next_instruction(test_id, method)
             d = _extract_data(data)
-            chunk_size = 4
+            bytes_yield = 0
             for r in range(0, len(d), chunk_size):
-                if r < 10:
+                if bytes_yield < limit:
```
tritone, line 656: I think if you want to preserve existing default behavior, you would have to choose a larger default limit than 4, correct? Maybe it doesn't matter too much.

testbench/common.py
```diff
@@ -644,27 +647,30 @@ def _extract_data(data):
...
-                # This exception is raised to abort the flow control. The
-                # connection has already been closed, causing the client to
-                # receive a "connection reset by peer" (or a similar error).
-                # The exception is useful in unit tests (where there is no
-                # socket to close), and stops the testbench from trying to
-                # complete a request that we intentionally aborted.
-                raise testbench.error.RestException(
-                    "Injected 'connection reset by peer' fault", 500
-                )
+                if bytes_yield >= limit:
```
tritone, line 661: This is just "else" I guess?

testbench/common.py
```diff
@@ -644,27 +647,30 @@ def _extract_data(data):
...
+                    if conn is not None:
+                        conn.close()
+                    # This exception is raised to abort the flow control. The
+                    # connection has already been closed, causing the client to
+                    # receive a "connection reset by peer" (or a similar error).
+                    # The exception is useful in unit tests (where there is no
+                    # socket to close), and stops the testbench from trying to
+                    # complete a request that we intentionally aborted.
+                    raise testbench.error.RestException(
+                        "Injected 'connection reset by peer' fault", 500
```
tritone, line 672: I guess technically this should say "Injected broken stream"

tritone (commented):
Couple little comments, LGTM overall. Thanks!

testbench/common.py
cojenco: Good catch, thanks!

testbench/common.py
cojenco: Done

testbench/common.py
cojenco: Yep, I considered that but given the updated configurability/logic, I think it's more straightforward to return a broken stream after 4 bytes as default behavior.

testbench/common.py
cojenco: hmmmm I guess so... does the server send back anything else after the payload? 

Is it possible to `return-broken-stream-after-256K` when downloading an exact 256K object with `chunk_size=256*1024` (the download is done in one chunk and then broken stream error)? I wrote two separate conditional checks intentionally for a test case like this.

testbench/common.py
tritone: That's fine, just wanted to point it out.

tritone (approved):

PR #284, opened 2022-02-16
feat: allow retry method storage.objects.download for testing

### LINKED ISSUE #283: Add Retry Test decorator method="storage.objects.download"
Introduce decorator with method `storage.objects.download` in the Retry Test API for testing purposes.

In retry conformance testing, we'd like to differentiate media operations from metadata get. This is to support complex download test cases for larger files.

cojenco:
Fixes #283

Adds and allows retry method `storage.objects.download` for testing purposes. 
This can help distinguish test cases that are performing media operations from metadata get.
- [X] Tests pass

### DIFF
--- a/testbench/rest_server.py
+++ b/testbench/rest_server.py
@@ -541,6 +541,7 @@ def object_delete(bucket_name, object_name):
 
 @gcs.route("/b/<bucket_name>/o/<path:object_name>")
 @retry_test(method="storage.objects.get")
+@retry_test(method="storage.objects.download")
 def object_get(bucket_name, object_name):
     blob = db.get_object(
         bucket_name,
--- a/testbench/rest_server.py
+++ b/testbench/rest_server.py
@@ -541,7 +541,6 @@ def object_delete(bucket_name, object_name):
 
 @gcs.route("/b/<bucket_name>/o/<path:object_name>")
 @retry_test(method="storage.objects.get")
-@retry_test(method="storage.objects.download")
 def object_get(bucket_name, object_name):
     blob = db.get_object(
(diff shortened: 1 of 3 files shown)


testbench/rest_server.py
```diff
@@ -541,6 +541,7 @@ def object_delete(bucket_name, object_name):
 
 @gcs.route("/b/<bucket_name>/o/<path:object_name>")
 @retry_test(method="storage.objects.get")
+@retry_test(method="storage.objects.download")
```
tritone, line 544: Just wanted to verify that you tested this out and it works to have multiple wrappers on a single method.

tritone (commented):
One question, otherwise LGTM

testbench/rest_server.py
cojenco: I did some testing and they pass running locally and live. Though I'm now concerned about possible chaining effects. I'll dig a bit deeper and change the patch. Thanks!

cojenco: I've found a better place to move the new download wrapper. When downloads perform, the two wrappers are still chained and will both run. I think the conditional checks in [`handle_retry_test_instruction`](https://github.com/googleapis/storage-testbench/blob/main/testbench/common.py#L699-L703) handle them well enough and therefore it works.

There's a better way to apply the new download wrapper by changing the `download_object_get` method. Though it requires changes in the conformance tests across all languages. Without the changes implemented, all tests would break.

cojenco:
Closing this PR for now as alternative solution is discussed in internal doc go/retry-conf-test-media-ops

PR #310, opened 2022-04-04
feat: update protos circa 2022-04-04

coryan:


gcs/bucket.py
```diff
@@ -85,7 +85,7 @@ def __preprocess_rest_ubla(cls, ubla):
     def __preprocess_rest_pap(cls, pap):
         pap = pap.upper()
         if pap == "UNSPECIFIED" or pap == "INHERITED":
-            return "PUBLIC_ACCESS_PREVENTION_UNSPECIFIED"
+            return "INHERITED"
```
frankyn, line 88: this is the only strange change; went from unspecified to inherited;

frankyn (approved):

PR #316, opened 2022-04-18
fix: add prefix support to list_buckets

### LINKED ISSUE #314: grpc & http: ListBuckets should support the Prefix field
The ListBucketsRequest has a [prefix field](https://github.com/googleapis/googleapis/blob/487c2cba96ea55d866ae19b464b7e424d38e8aa4/google/storage/v2/storage.proto#L359) (so does [JSON](https://cloud.google.com/storage/docs/json_api/v1/buckets/list#parameters)), which the server doesn't currently support in both HTTP and gRPC. 

It would be nice to support `prefix` as I'd like to test that my iterator logic works as expected when additional fields (beyond pagination fields) are set.

noahdietz:
Fixes #314 

- [x] Tests pass
- [ ] ~Appropriate changes to README are included in PR~ **N/A**

### DIFF
--- a/testbench/database.py
+++ b/testbench/database.py
@@ -97,11 +97,20 @@ class Database:
             self._objects[bucket.metadata.name] = {}
             self._live_generations[bucket.metadata.name] = {}
 
-    def list_bucket(self, project_id, context):
+    def list_bucket(self, project_id, prefix, context):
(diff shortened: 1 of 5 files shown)


cojenco (approved):
LGTM, thanks!

PR #317, opened 2022-04-22
fix: prepend bucket name prefix for both transports

noahdietz:
I thought only grpc-created buckets had the bucket name prefixed as a resource, but both do. This meant that http List Bucket calls with Prefix never worked unless the full project resource name was also part of the prefix.

Fixes the fix for #314.


frankyn (approved):

tritone (approved):
nice catch!

PR #322, opened 2022-05-01
fix: decompressive transcoding improvements

### LINKED ISSUE #321: Review implementation of decompressive transcoding
I am not sure the implementation of decompressive transcoding works as it should.  It seems to decompress the data when uploaded:

https://github.com/googleapis/storage-testbench/blob/465f85e604f2ee526e10a774b4f8ead6c667c386/testbench/handle_gzip.py#L29-L34

But that probably won't work with `uploadType=multipart`, where the `Content-Type` is `multipart/related;`.  I think we need to decompress the results on the fly, and disable decompression if the `Accept-Encoding: gzip` flag is enabled.  We also need to store the checksums of the compressed data, which suggests (again) that decompressing the request is not the right idea.

coryan:
Store the data in compressed form (if applicable), I believe this is a
better approach as it:

- Makes the `size` and checksum attributes match the behavior in the
  service.
- It enables support for `Accept-Encoding`, we can return compressed
  data.
- It makes the gRPC case match the service behavior, i.e., always return
  compressed data.

Fixes #321


tritone (approved):
Very nice improvement, thank you!

Issue #306, opened 2022-04-01
Remove dependency on `httpbin`


coryan:
The Cloud C++ team uses an `httpbin` endpoint to test some of their low-level libcurl wrappers.

https://github.com/googleapis/storage-testbench/blob/c80ed7604824428a7b8269504ec948a62b445f5d/testbench/rest_server.py#L1047

We need to break this dependency because the bits in `httpbin` are rotting and nobody is replacing them with new shinny bits.  I did a quick scan and this is a first attempt at listing all the routes we use:

- [ ] `/get`
- [ ] `/get?foo=foo-value`
- [ ] `/post`
- [ ] `/put`
- [ ] `/stream/<line-count>`
- [ ] `/response-headers`
- [ ] `/status/<N>`
- [ ] `/response-headers?param=value`
- [ ] `/headers`
- [ ] `/post?foo=bar`
- [ ] `/delete`
- [ ] `/patch`
- [ ] `/anything`

PR #323, opened 2022-05-02
test: add httpbin tests

coryan:
I am planning to remove the dependency on the `httpbin` module. This
adds some tests to verify the replacement largely works as the current
implementation.

Part of the work for #306

### DIFF
--- a/tests/test_echo.py
+++ b/tests/test_echo.py
@@ -1,0 +1,403 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


tests/test_echo.py
```diff
@@ -0,0 +1,403 @@
...
+
+from testbench import rest_server
+
+
+class TestEcho(unittest.TestCase):
+    def setUp(self):
+        self.client = rest_server.server.test_client()
+
+    def test_delete(self):
+        expected_args = {"arg1": "value1", "arg2": "value2"}
```
tritone, line 32: Maybe worth factoring out some of this duplicated data across tests? Up to you either way, though

tritone (approved):

tests/test_echo.py
coryan: I am Okay with the dups, thanks.

PR #324, opened 2022-05-02
fix: default to resumable uploadType if not passed in query param, return location as X-Good-Upload-URL header

bajajneha27:
1. From Ruby library, we do not send `uploadType` as a query parameter, the first time we call upload object API. Since, we default to `resumable` upload type anyway at the end, I thought it'd be a good idea to set it to default if not passed in query param.
2. Ruby client lib also looks for header `X-Goog-Upload-URL`  to make next API call.

- [ ] Tests pass
- [ ] Appropriate changes to README are included in PR

### DIFF
--- a/testbench/rest_server.py
+++ b/testbench/rest_server.py
@@ -879,7 +879,7 @@ upload.register_error_handler(Exception, testbench.error.RestException.handler)
 def object_insert(bucket_name):
     db.insert_test_bucket()
     bucket = db.get_bucket(bucket_name, None).metadata
-    upload_type = flask.request.args.get("uploadType")
+    upload_type = flask.request.args.get("uploadType", "resumable")
     if upload_type is None:
         testbench.error.missing("uploadType", None)
     elif upload_type not in {"multipart", "media", "resumable"}:
@@ -889,6 +889,7 @@ def object_insert(bucket_name):
         db.insert_upload(upload)
         response = flask.make_response("")
         response.headers["Location"] = upload.location
+        response.headers["X-Goog-Upload-URL"] = upload.location
         return response
     blob, projection = None, ""
     if upload_type == "media":
--- a/tests/test_testbench_object_upload.py
+++ b/tests/test_testbench_object_upload.py
@@ -464,15 +464,16 @@ class TestTestbenchObjectUpload(unittest.TestCase):
         query_rest.pop("owner")
         self.assertEqual(insert_rest, query_rest)
 
-    def test_upload_validate_upload_type(self):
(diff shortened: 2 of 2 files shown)


testbench/rest_server.py
```diff
@@ -889,6 +889,7 @@ def object_insert(bucket_name):
         db.insert_upload(upload)
         response = flask.make_response("")
         response.headers["Location"] = upload.location
+        response.headers["X-Goog-Upload-URL"] = upload.location
```
coryan, line 892: Do you want to add a test to verify this works?

coryan: Also, the documented header is `Location`:

https://cloud.google.com/storage/docs/performing-resumable-uploads

The `X-Goog-Upload-URL` seems to be undocumented, and undocumented behavior may change at any time.

testbench/rest_server.py
```diff
@@ -879,7 +879,7 @@ def download_object_get(bucket_name, object_name):
 def object_insert(bucket_name):
     db.insert_test_bucket()
     bucket = db.get_bucket(bucket_name, None).metadata
-    upload_type = flask.request.args.get("uploadType")
+    upload_type = flask.request.args.get("uploadType", "resumable")
```
coryan, line 882: If the Ruby library depends on this, it is depending on undocumented behavior.  The `uploadType` parameter is a **required** parameter:

https://cloud.google.com/storage/docs/json_api/v1/objects/insert#parameters

coryan (approved):
I think you should split this into two separate PRs.  And you need to adjust the tests in any case.

Finally, it seems Ruby is depending on undocumented behavior.  The "Right Thing"[tm] would be to fix the Ruby client to only use the documented behavior of the service.

testbench/rest_server.py
tritone: I'm a little confused about this. I tested out creating a resumable upload session and I actually don't see `x-goog-upload-url` in the response at all, only `location`.

However I do see it documented for other random Google APIs, e.g. https://developers.google.com/photos/library/guides/resumable-uploads#:~:text=The%20header%20field%20X%2DGoog%2DUpload%2DURL%20contains%20a,use%20it%20for%20subsequent%20requests.

testbench/rest_server.py
tritone: I tried this as well... interestingly the JSON API seems to accept requests without UploadType, but it defaults to a simple upload rather than resumable.

bajajneha27:
As discussed, we'll be making changes in the Ruby library itself. I'll go ahead and close this one. Thanks everyone :)

Issue #338, opened 2022-05-14
Many fields not filled in `*AccessControl: *` operations


coryan:
The testbench stores its data in proto format.  For some resources, including `BucketAccessControl` the protos elide a number of redundant fields.  The testbench carefully fills back these fields when returning the data via `Bucket: *` REST operations, for example here:

https://github.com/googleapis/storage-testbench/blob/e74c1205dd91e3f4cd00aafc723d7f7eca916880/gcs/bucket.py#L290-L298

However, we neglect to do the same when returning these objects via the `BucketAccessControl: *` RPCs:

https://github.com/googleapis/storage-testbench/blob/e74c1205dd91e3f4cd00aafc723d7f7eca916880/testbench/rest_server.py#L301-L302

That makes some of the tests hard to write, as you have to ignore some fields depending on what RPC you used to get the resource.

We have these problems with:

- [x] `BucketAccessControl: *`
- [x] `ObjectAccessControl: *`
- [ ] `DefaultObjectAccessControl: *`

PR #342, opened 2022-05-18
feat: improve `BucketAcl: *` responses

coryan:
Add attributes not present in the protos.

Part of the work for #338

### DIFF
--- a/tests/test_proto2rest.py
+++ b/tests/test_proto2rest.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright 2020 Google LLC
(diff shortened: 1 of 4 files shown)


tests/test_proto2rest.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+#
+# Copyright 2020 Google LLC
```
tritone, line 3: Update license year

tritone (approved):

tests/test_proto2rest.py
coryan: Fixed, thanks.

PR #343, opened 2022-05-18
refactor: split object proto2rest functions

coryan:
Move the functions to enrich the REST representation of objects to a
separate file.  I am plannning to further refactor this code to improve the
values returned by `ObjectAccessControl: *` RPCs.

Part of the work for #338

### DIFF
--- a/tests/test_proto2rest.py
+++ b/tests/test_proto2rest.py
@@ -52,6 +52,117 @@ class TestProto2Rest(unittest.TestCase):
         self.assertEqual(actual, {**actual, **expected})
         self.assertIn("etag", actual)
 
+    def test_object_as_rest(self):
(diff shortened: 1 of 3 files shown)


tests/test_proto2rest.py
```diff
@@ -52,6 +52,117 @@ def test_bucket_access_control_as_rest(self):
         self.assertEqual(actual, {**actual, **expected})
         self.assertIn("etag", actual)
```
dbolduc, line 54: nit: second newline?

dbolduc (approved):

tests/test_proto2rest.py
coryan: https://peps.python.org/pep-0008/#blank-lines

PR #345, opened 2022-05-19
feat: enrich `ObjectAccessControl` REST responses

coryan:
Always use the helper functions in `proto2rest` when converting
`storage_pb2.ObjectAccessControl` to their REST representation.  Needed
some light refactoring in the `proto2rest` module to make this work.

Part of the work for #338

### DIFF
--- a/testbench/proto2rest.py
+++ b/testbench/proto2rest.py
@@ -28,6 +28,7 @@ This module contains a number of helper functions to perform these transformatio
 
 import base64
 import hashlib
+from os import access
 import struct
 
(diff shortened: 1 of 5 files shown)


testbench/proto2rest.py
```diff
@@ -28,6 +28,7 @@
 
 import base64
 import hashlib
+from os import access
```
devbww, line 31: This looks strange.

coryan: Indeed, fixed.

devbww (approved):

PR #371, opened 2022-07-13
chore: update protos circa 2022-07-03

coryan:


coryan:
FWIW, the errors are coverage changes in the protbuf-generated code, we always ignore those.

Issue #331, opened 2022-05-04
Validate values in `Range` header


coryan:
The testbench is not validating that the `Range` header values are in the right range for the object contents.

It should return a `416` error (`REQUEST RANGE NOT SATISFIABLE`).

PR #397, opened 2022-08-26
feat: support status codes 206 and 416 for range downloads

### LINKED ISSUE #396: Return 206 for valid range requests
Add support for the testbench to return [`206 Partial Content`](https://datatracker.ietf.org/doc/html/rfc7233#section-4.1) status code for valid range download requests

cojenco:
This adds validation to range download request headers
- testbench returns `206 Partial Content` for valid range download requests
- testbench returns `416 Requested Range Not Satisfiable` for invalid range headers

Fixes #396 and some of #331

- [x] Tests pass

### DIFF
--- a/gcs/object.py
+++ b/gcs/object.py
@@ -376,6 +376,10 @@ class Object:
         begin, end, length, response_payload = self._download_range(
             request, response_payload
         )
+        # Return 416 if the requested range cannot be satisfied.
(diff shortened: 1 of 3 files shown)


coryan (approved):

tritone (approved):
Thanks for the fix Cathy!

PR #401, opened 2022-09-09
feat: support gRPC metadata echo

varunramc:
gRPC server can now optionally echo request metadata for all RPC methods. Echoed headers are prefixed with 'x-req-'. The feature is controlled via an 'echo-metadata' query parameter to '/start_grpc'.

### DIFF
--- a/testbench/grpc_server.py
+++ b/testbench/grpc_server.py
@@ -98,28 +98,46 @@ def _logging_method_decorator(function):
 
     return decorated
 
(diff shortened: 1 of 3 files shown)


coryan (changes requested):
Looks great, thanks!  Please format using black==22.3.0:

https://github.com/googleapis/storage-testbench/runs/8261228010?check_suite_focus=true

coryan (approved):
Thanks!

PR #403, opened 2022-09-09
feat: support request header X-Guploader-No-308

### LINKED ISSUE #402: Support request header `X-GUploader-No-308`
Add support for such clients as apiary-go that add a `X-GUploader-No-308: yes` request header

If `X-GUploader-No-308: yes` is included in the resumable upload request headers, instead of returning `308: resume incomplete`, [return `200` to satisfy the client and add a `X-HTTP-Status-Code-Override: 308` response header](go/scotty-faq#my-client-received-an-http-308-response-with-no-location-header)

cojenco:
Fixes #402

This adds support for `X-GUploader-No-308: yes`  - if included in resumable upload request headers, instead of returning 308: resume incomplete, [return 200 to satisfy the client and add a X-HTTP-Status-Code-Override: 308 response header](https://github.com/googleapis/storage-testbench/issues/go/scotty-faq#my-client-received-an-http-308-response-with-no-location-header)

- [x] Tests pass

### DIFF
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -241,13 +241,13 @@ class Upload(types.SimpleNamespace):
             upload.metadata.metadata["x_emulator_no_md5"] = "true"
         return upload, is_resumable
 
-    def resumable_status_rest(self, no_308=False):
+    def resumable_status_rest(self, override_308=False):
         response = flask.make_response()
         if len(self.media) > 1 and not self.complete:
             response.headers["Range"] = "bytes=0-%d" % (len(self.media) - 1)
         response.status_code = 308
 
-        if no_308:
+        if override_308:
             response.headers["X-Http-Status-Code-Override"] = "308"
             response.status_code = 200
         return response
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -241,9 +241,13 @@ class Upload(types.SimpleNamespace):
             upload.metadata.metadata["x_emulator_no_md5"] = "true"
         return upload, is_resumable
 
-    def resumable_status_rest(self):
+    def resumable_status_rest(self, no_308=False):
         response = flask.make_response()
         if len(self.media) > 1 and not self.complete:
             response.headers["Range"] = "bytes=0-%d" % (len(self.media) - 1)
         response.status_code = 308
+
+        if no_308:
+            response.headers["X-Http-Status-Code-Override"] = "308"
+            response.status_code = 200
         return response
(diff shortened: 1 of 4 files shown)


gcs/upload.py
```diff
@@ -241,9 +241,13 @@ def init_write_object_grpc(cls, db, request_iterator, context):
             upload.metadata.metadata["x_emulator_no_md5"] = "true"
         return upload, is_resumable
 
-    def resumable_status_rest(self):
+    def resumable_status_rest(self, no_308=False):
```
coryan, line 244: `no_Foo = False` is a code smell. Why two negations to express what we want? Consider `use_308=True`

And does this have to have a default value?  It seems that we would want to change all call sites to pass the right parameter in?

testbench/rest_server.py
```diff
@@ -901,9 +901,15 @@ def handle_gzip_compressed_request():
 @upload.route("/b/<bucket_name>/o", methods=["POST"])
 @retry_test(method="storage.objects.insert")
 def object_insert(bucket_name):
+    # GCS supports "POST" calls for uploading data with an upload_id
```
coryan, line 904: This seems like a good change, but unrelated to this PR?  Or maybe the PR description needs to be more general? e.g. improve resumable uploads?

coryan: It also looks like this code is tested, but not by a new explicit test, I wonder how that worked before.

coryan (changes requested):
The double negative is a deal breaker for me.  The other comments are more questions rather than requests for changes.

testbench/rest_server.py
cojenco: Good call out. Splitting this into a separate PR #404. Not sure how test coverage was done here 🤷🏻‍♀️ but updated tests for the new changes.

gcs/upload.py
cojenco: Well noted! With that in mind, I'd like to use `override_308=False`

Considering how GCS works or at least how it's documented publicly, I think it makes sense to have a default value `override_308=False`. IIUC, `308 Resume Incomplete` is the default GCS response; `308` is overriden by `200` under resume incomplete only when the request header `X-Guploader-No-308=yes` is included. Does that seem reasonable?

cojenco: Updated with proposed modifications for now. Happy to make further changes!

coryan: Still weird that the default is to say "please don't do X" (with `X==override_308`), but I think your argument about consistency with the GCS headers is compelling.  LGTM

coryan (approved):

tritone (approved):

PR #417, opened 2022-10-07
feat: support autoclass toggle time

coryan:


cojenco (approved):
LGTM, thanks for adding this! btw this supports v1 and v2 autoclass ;)

PR #442, opened 2022-12-21
chore: reduce update frequency for Dockerfile

coryan:


renovate.json
```diff
@@ -6,7 +6,7 @@
     {
       "matchManagers": ["dockerfile"],
       "matchPackageNames": ["python"],
-      "matchCurrentVersion": "3.10",
+      "matchCurrentVersion": "3.11",
```
dbolduc, line 9: how does this "reduce update frequency" ?

why is this PR separate from #443 ?

coryan: It reduces the update frequency because the Dockerfile used to refer to Python 3.10, but now refers to Python 3.11 so the rule to lower the frequency of updates no longer matches.

They are separate because I noticed the lack of testing after I had sent this PR.

dbolduc (approved):

PR #449, opened 2023-01-05
feat: Support both `waitress` and `gunicorn` servers

anuragsrivstv:
As gunicorn is not supported on windows , these changes are to dynamically switch between waitress and gunicorn, so that waitress can be used as a server whenever the testbench is on windows , otherwise gunicorn.

Also added new job in build.yaml to test storage test bench on windows.

- [x] Tests pass

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,3 +37,35 @@ jobs:
           flags: unittests
           name: codecov-umbrella
           verbose: true
+  windows-tests:
+    runs-on: windows-2022
+    strategy:
+      matrix:
+        python-version: ["3.11"]
+    name: Unit Tests On Windows Python ${{ matrix.python-version }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8 pytest pytest-cov
+          pip install -e .
+      - name: Run Unit Tests with Coverage
+        env:
+          PYTHONPATH: "."
+        run: |
+          pytest --cov=./ --cov-report=xml
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          env_vars: OS,PYTHON
+          fail_ci_if_error: true
+          files: ./coverage.xml
+          flags: unittests
+          name: codecov-umbrella
+          verbose: true
+          
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,9 +20,8 @@ COPY . /opt/storage-testbench/
 
 RUN python3 -m pip install -e .
 
-CMD ["gunicorn", \
-      "--bind", "0.0.0.0:9000", \
-      "--worker-class", "sync", \
-      "--threads", "10", \
-      "--access-logfile", "-", \
-      "testbench:run()"]
+CMD ["python3", \
+      "testbench_run.py", \
+      "0.0.0.0", \
+      "9000", \
+      "10"]
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -34,6 +34,7 @@ import scalpl
 
 from google.storage.v2 import storage_pb2
 import testbench
+import platform
 
 re_remove_index = re.compile(r"\[\d+\]+|^[0-9]+")
 retry_return_error_code = re.compile(r"return-([0-9]+)$")
@@ -649,7 +650,41 @@ def _extract_headers(response):
     return dict()
 
 
-def __get_streamer_response_fn(database, method, conn, test_id, limit=4, chunk_size=4):
+def _using_waitress(request):
+    return request.environ.get("SERVER_SOFTWARE", "").startswith("waitress")
+
+
+def _get_socket(request):
+    if _using_waitress(request):
+        channel = request.environ.get("waitress.channel", None)
+        if channel is None:
+            return None
+        return channel.socket
+    return request.environ.get("gunicorn.socket", None)
+
+
+def _make_closer(request):
+    def _null_closer():
+        pass
+
+    if _using_waitress(request):
+        channel = request.environ.get("waitress.channel", None)
+
+        def _closer():
+            channel.close()
+
+        return _null_closer if channel is None else _closer
+    socket = request.environ.get("gunicorn.socket", None)
+
+    def _closer():
+        socket.close()
+
+    return _null_closer if socket is None else _closer
+
+
+def __get_streamer_response_fn(
+    database, method, socket_closer, test_id, limit=4, chunk_size=4
+):
     def response_handler(data):
         def streamer():
             d = _extract_data(data)
@@ -662,8 +697,7 @@ def __get_streamer_response_fn(database, method, conn, test_id, limit=4, chunk_s
                     yield d[r:chunk_end]
                 if bytes_yield >= limit:
                     database.dequeue_next_instruction(test_id, method)
-                    if conn is not None:
-                        conn.close()
+                    socket_closer()
                     # This exception is raised to abort the flow control. The
                     # connection has already been closed, causing the client to
                     # receive a "broken stream" or "connection reset by peer" error.
@@ -698,7 +732,7 @@ def __get_limit_response_fn(database, upload_id, test_id, method, limit):
     return limited_response_fn
 
 
-def handle_retry_test_instruction(database, request, method):
+def handle_retry_test_instruction(database, request, socket_closer, method):
     upload_id = request.args.get("upload_id", None)
     test_id = request.headers.get("x-retry-test-id", None)
     if not test_id or not database.has_instructions_retry_test(test_id, method):
@@ -724,12 +758,13 @@ def handle_retry_test_instruction(database, request, method):
         items = list(retry_connection_matches.groups())
         if items[0] == "reset-connection":
             database.dequeue_next_instruction(test_id, method)
-            fd = request.environ.get("gunicorn.socket", None)
-            if fd is not None:
-                fd.setsockopt(
+            sock = _get_socket(request)
+            if sock is not None:
+                sock.setsockopt(
                     socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0)
                 )
-                fd.close()
+                socket_closer()
+
             # This exception is raised to abort the flow control. The connection
             # has already been closed, causing the client to receive a "connection
             # reset by peer" (or a similar error). The exception is useful in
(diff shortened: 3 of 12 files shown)


testbench_waitress.py
```diff
@@ -0,0 +1,55 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from waitress.channel import HTTPChannel
+from waitress.task import WSGITask
+from waitress import create_server
+
+# We need request sockets of waitress to produce "broken-stream" and "return-reset-connection"
+# error. If we are going to update waitress vesion in setup.py then code in this file may also needs to be verified and updated accordingly.
```
coryan, line 20: Remove "If we are going ..." that is always true.  Tests need adjustment if the code / dependencies chance in incompatible way.

coryan (commented):
I think the tense in the PR title is off. You do not want to say what you did, but what the PR **does**.  e.g. "feat: support both `waitress` and `gunicorn` servers"

.github/workflows/build.yaml
```diff
@@ -37,3 +37,34 @@ jobs:
           flags: unittests
           name: codecov-umbrella
           verbose: true
+  windows-tests:
+    runs-on: windows-latest
```
coryan, line 41: I tend to avoid `*-latest` things because our builds will break when GitHub decides to update.  Pin to the latest version explicitly and have something like renovatebot update when needed is better.

testbench/common.py
```diff
@@ -701,6 +699,9 @@ def limited_response_fn(data):
 def handle_retry_test_instruction(database, request, method):
     upload_id = request.args.get("upload_id", None)
     test_id = request.headers.get("x-retry-test-id", None)
+    running_on_windows = False  # Default flag value.
```
coryan, line 702: Use

```py
  running_on_windows = (platform.system().lower() == "windows")
```

Or better yet, refactor that to a member function.

testbench/common.py
```diff
@@ -724,12 +725,22 @@ def handle_retry_test_instruction(database, request, method):
...
             database.dequeue_next_instruction(test_id, method)
-            fd = request.environ.get("gunicorn.socket", None)
-            if fd is not None:
-                fd.setsockopt(
-                    socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0)
-                )
-                fd.close()
+            fd = None
+            channel = None
+            if running_on_windows:
```
coryan, line 730: But this is not about whether you are running on Windows, is it?  It is about whether you are using `waitress`.  For example, if we decided to use `waitress` on Linux this test would be all wrong?  Please fix to test for the presence of `waitress`.

It seems that `request.environ.get("SERVER_SOFTWARE")` can be used for this purpose.

testbench/common.py
```diff
@@ -724,12 +725,22 @@ def handle_retry_test_instruction(database, request, method):
...
+                channel = request.environ.get("waitress.channel", None)
+                if channel is not None:
```
coryan, line 732: I suggest you hide all this complexity behind a function that returns the socket.  Because that is what you really need.

testbench/common.py
```diff
@@ -739,17 +750,25 @@ def handle_retry_test_instruction(database, request, method):
                 "Injected 'connection reset by peer' fault", 500
             )
         elif items[0] == "broken-stream":
-            conn = request.environ.get("gunicorn.socket", None)
-            return __get_streamer_response_fn(database, method, conn, test_id)
+            if running_on_windows:
+                conn_or_channel = request.environ.get("waitress.channel", None)
```
coryan, line 754: I think you could do something like:

```
  def _using_gunicorn(self, request):
    # TODO(...) - double check this, I just read the documentation and guessed at the value.
    return request.environ.get("SEREVER_SOFTWARE", "").startswith("gunicorn")

  def _get_socket(self, request):
    if self._using_gunicorn(request):
      return request.environ.get("gunicorn.socket", None)
    channel =  request.environ.get("waitress.channel", None)
    if channel is None: return channel
    return channel.socket
```

coryan (changes requested):
Sorry for sending two reviews, I clicked on the wrong button.

coryan (changes requested):

testbench_run.py
```diff
@@ -0,0 +1,56 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import platform
+import waitress
+import testbench
+import logging
+import subprocess
+from testbench_waitress import testbench_create_server
```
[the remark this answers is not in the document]
anuragsrivstv: Yes. Added unit tests.

anuragsrivstv:
Unit Test on python version <= 3.8 has failed. Looking into it.

.github/workflows/build.yaml
```diff
@@ -37,3 +37,35 @@ jobs:
...
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          env_vars: OS,PYTHON
+          fail_ci_if_error: true
+          files: ./coverage.xml
+          flags: unittests
+          name: codecov-umbrella
+          verbose: true
+
```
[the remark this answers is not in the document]
coryan: This is still broken.

Dockerfile
```diff
@@ -20,9 +20,9 @@ COPY . /opt/storage-testbench/
...
-      "--worker-class", "sync", \
-      "--threads", "10", \
-      "--access-logfile", "-", \
-      "testbench:run()"]
+CMD ["python3", \
+      "testbench_run.py", \
+      "0.0.0.0", \
+      "9000", \
+      "10"]
+
```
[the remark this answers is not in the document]
coryan: Ditto.

testbench/common.py
```diff
@@ -662,8 +697,11 @@ def streamer():
                     yield d[r:chunk_end]
                 if bytes_yield >= limit:
                     database.dequeue_next_instruction(test_id, method)
-                    if conn is not None:
-                        conn.close()
+                    # Accessing request object in unit test will give "working out of request context" exception. Ignoring these exception.
+                    try:
+                        _make_closer(request)
```
coryan, line 702: `_make_closer(request)` returns a function that should be called to have any effect?

testbench/common.py
```diff
@@ -724,12 +762,13 @@ def handle_retry_test_instruction(database, request, method):
...
-            fd = request.environ.get("gunicorn.socket", None)
-            if fd is not None:
-                fd.setsockopt(
+            sock = _get_socket(request)
+            if sock is not None:
+                sock.setsockopt(
                     socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0)
                 )
-                fd.close()
+                _make_closer(request)
```
coryan, line 770: Same here.  You need to do something with the result of this function?

testbench_waitress.py
```diff
@@ -0,0 +1,54 @@
...
+
+class testbench_HTTPChannel(HTTPChannel):
+    task_class = testbench_WSGITask
+
+
+def testbench_create_server(
+    application, map=None, _start=True, _sock=None, _dispatcher=None, **kw
+):
+    if map is None:
+        map = {}
```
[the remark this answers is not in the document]
coryan: Did you make any action?

tests/test_testbench_continue_after_fault_injection.py
```diff
@@ -177,7 +178,7 @@ def test_repeated_error_after_256K_faults_by_header(self):
         self.assertEqual(response.status_code, 200)
 
         # Verify we get the expected error when sending several requests
-        for _ in range(0, 10):
+        for _ in range(0, 4):
```
coryan, line 181: I am not sure `10` was the right value, but I am not sure why `4` is the right value either.  Can you add a comment explaining why?  If not, just leave the old value, less explaining to do that way 😄

coryan: Looking at it, I suspect `10` was related to the number of processes / threads started in line 42.

testbench/common.py
```diff
@@ -662,8 +697,11 @@ def streamer():
                     yield d[r:chunk_end]
                 if bytes_yield >= limit:
                     database.dequeue_next_instruction(test_id, method)
-                    if conn is not None:
-                        conn.close()
+                    # Accessing request object in unit test will give "working out of request context" exception. Ignoring these exception.
```
coryan, line 700: The answer to that may be to call `_make_closer()` in the request context, and pass the result (the created "closer" function) to `__get_streamer_response_fn()`.

coryan (changes requested):
It seems that some comments are still unaddressed, and the code may does not work as I would expect.

testbench_waitress.py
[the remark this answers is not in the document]
anuragsrivstv: No, as when I removed default it gave error while starting testbench. 

So Basically this default value( i.e. `None`) of map is not unit tested because in unit test ( 
 `test_created_wsgi_server_should_have_testbench_httpChannel ` ) I need to initialize and provide `map` to testbench_create_server(..) so that I can assert populated wsgi server in the `map`.  
But still we need to keep  `None` as default value of map because in ` waitress\__init__.py ` server ( which is return of testbench_create_server ) is invoked only with application( ` server = _server(app, **kw) ` )

Please let me know if you want me to try something else.

tests/test_testbench_continue_after_fault_injection.py
anuragsrivstv: Sure. I changed value to 4 because on testing it with waitress, it was failing intermittently for value greater than 4. At this point I'm not sure that this is failing due to waitress , or due to windows or is it a behavior of code for error_after_256K  instructions which is surfaced due to waitress or something else.

So right now I have reverted the loop back to 10 times iteration, but have skipped it for windows currently and raised an issue( #454) to track it. Please let me know if it there is a better approach to handle it.

testbench/common.py
anuragsrivstv: > It seems that some comments are still unaddressed, and the code may does not work as I would expect.

I've addressed all comments now. Please let me know if any of the change is not inline with what you have suggested,

testbench_waitress.py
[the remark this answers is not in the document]
coryan: Ack.  Maybe a comment explaining that is only intended for testing?

coryan (approved):
A final suggestion for you to consider.

testbench_waitress.py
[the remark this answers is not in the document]
anuragsrivstv: Or should we add `# pragma: no cover ` to exclude this line from coverage ?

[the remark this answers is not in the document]
coryan: I think it is more important to say why than to ignore the coverage problem.  Your call.

coryan (approved):

testbench_waitress.py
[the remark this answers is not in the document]
anuragsrivstv: Got it. Added the comment. Please have a look, after it I'll squash the commits into one and merge.

coryan (approved):

PR #455, opened 2023-01-18
fix: update Object.rest_media to set content-* headers from object metadata if present

BenWhitehead:


gcs/object.py
```diff
@@ -495,6 +495,20 @@ def streamer():
         headers["x-goog-metageneration"] = self.metadata.metageneration
         headers["x-goog-storage-class"] = self.metadata.storage_class
 
+        if self.metadata.content_type:
+            headers["Content-Type"] = self.metadata.content_type
```
coryan, line 499: Can you remind me how this interacts with decompressive transcoding?  What is supposed to go here when the data is decompressed?  What about when it is not?

BenWhitehead: My understanding (primarily from the standpoint of http, less storage specific), is outlined in the following table:

| object metadata                                       | request                                 | expected response                                                                                                                                                                                |
|-------------------------------------------------------|-----------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `{contentType:"text/plain"}`                          | `Accept: *`                             | `Content-Type: text/plain`<br/> + raw bytes, without any transform                                                                                                                               |
| `{contentType:"text/plain", contentEncoding: "gzip"}` | `Accept: *`                             | `Content-Type: text/plain`<br/> + server decompressed bytes                                                                                                                                      |
| `{contentType:"text/plain", contentEncoding: "gzip"}` | `Accept: *`<br/>`Accept-Encoding: gzip` | `Content-Type: text/plain`<br/>`Content-Encoding: gzip`<br/> + raw bytes, without any transform                                                                                                  |
| `{contentEncoding: "gzip"}`                           | `Accept: *`<br/>`Accept-Encoding: gzip` | `Content-Type: application/octet-stream`<br/>`Content-Encoding: gzip`<br/> + raw bytes, without any transform                                                                                    |
| `{}`                                                  | `Accept: *`<br/>`Accept-Encoding: gzip` | `Content-Type: application/octet-stream`<br/> + raw bytes, without any transform<br/>OR<br/>`Content-Type: application/octet-stream`<br/>`Content-Encoding: gzip`<br/> + server compressed bytes |
| `{}`                                                  | `Accept: *`                             | `Content-Type: application/octet-stream`<br/> + raw bytes, without any transform                                                                                                                 |

coryan: Thanks.

coryan (approved):
LGTM. Seems like we have a flake, maybe just re-run the build, but now I wonder why is that test non-deterministic.... sigh.

gcs/object.py
```diff
@@ -368,9 +368,11 @@ def _download_range(self, request, response_payload):
         return begin, end, length, response_payload
 
     def rest_media(self, request, delay=time.sleep):
+        print(self.metadata)
```
coryan, line 371: debug?

coryan (changes requested):

gcs/object.py
BenWhitehead: yup, cleaned it up. Sorry about that.

gcs/object.py
```diff
@@ -488,15 +490,34 @@ def streamer():
...
         headers["x-goog-storage-class"] = self.metadata.storage_class
 
+        if self.metadata.content_type:
+            headers["Content-Type"] = self.metadata.content_type
+        else:
+            # GCS json defaults to application/octet-stream if the object
+            # doesn't specify its content-type
+            headers["Content-Type"] = "application/octet-stream"
+
+        headers["x-goog-stored-content-length"] = self.metadata.size
```
cojenco, line 504: Thanks for adding this header :)

gcs/object.py
BenWhitehead: For posterity (courtesy of @cojenco) https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding dives into the header expectations from GCS's standpoint when Decompressive transcoding is applied.

She also helped me better understand the various pre-computation at play in rest_media such that we can selectively include `Content-Encoding` only when Decompressive transcoding is not used.

coryan (approved):

cojenco (approved):

PR #481, opened 2023-03-23
fix(gRPC): validate ReadObject.read_offset

laramiel:
When a GCS request is made, read_offset is compared with the object size, and if it's too large, an error is returned.


coryan:
FYI, conventional commits are of the form:

```
keyword(component): short statement
```

Where `keyword` is selected from a short list: `fix`, `doc`, `feat`, `test`, `ci` are common. We also use `impl`.  `component` is supposed to be a thing in the library / program.  We don't use that a lot in the testbench, and the short statement should say what it fixes.  The whole line must be less than 50 characters.

coryan (approved):

PR #526, opened 2023-08-17
feat: initial scaffolding to add Retry Test API in gRPC

cojenco:
-  Initial scaffolding for new retry test handler in the gRPC server 
    - enable method `GetBucket`
    - error-codes-matches (e.g. return-503) 
    - retry_connection_matches (e.g. return-reset-connection)

- Add new field `"transport": "JSON"` to the Retry Test Resource
    - trigger retry test cases for specified transport(s), returns 501 Unimplemented for operations not yet supported 
    - validate and distinguish retry test instructions for specified transport

- Add error code mapping between HTTP and gRPC

- [x] Tests pass
- Appropriate changes to README will be added when the support is completed

### DIFF
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -755,11 +755,27 @@ def grpc_handle_retry_test_instruction(database, request, context, method):
         database.dequeue_next_instruction(test_id, method)
         items = list(error_code_matches.groups())
         rest_code = items[0]
-        # TODO: Map error code between http and gRPC
-        grpc_code = StatusCode.UNAVAILABLE
+        grpc_code = map_closest_http_to_grpc(rest_code)
+        error_message = {
+            "error": {"message": "Retry Test: Caused a {}".format(grpc_code)}
+        }
         print("### injecting error ###")
         testbench.error.inject_error(context, rest_code, grpc_code, msg="")
     return __get_default_response_fn
+
+
+def map_closest_http_to_grpc(http_code):
+    # Only map the status codes that are used in tests and listed in
+    # the README, to avoid error-prone code conversions.
+    status_map = {
+        "400": StatusCode.INVALID_ARGUMENT,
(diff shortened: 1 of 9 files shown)


coryan (approved):
Candidly this is going over my head, but if you need a rubber stamp I am happy to grant it.

testbench/common.py
```diff
@@ -737,10 +738,55 @@ def limited_response_fn(data):
     return limited_response_fn
 
 
+def grpc_handle_retry_test_instruction(database, request, context, method):
+    test_id = None
+    if context is not None:
+        if hasattr(context, "invocation_metadata") and isinstance(
```
andrewsg, line 744: If context isn't these things, should there be a ValueError?

testbench/common.py
```diff
@@ -737,10 +738,55 @@ def limited_response_fn(data):
...
+            context.invocation_metadata(), Iterable  # Handle mocks in tests
+        ):
+            metadata = context.invocation_metadata()
+            if not isinstance(metadata, dict):
+                metadata = dict(metadata)
```
andrewsg, line 749: What type are you expecting metadata to be here if not a dict already? Most iterables can't be cast to dict.

andrewsg (changes requested):

testbench/common.py
cojenco: If the context isn't these things, the retry handler is expected to not do anything, and simply let the request pass through by returning the default response.

testbench/common.py
cojenco: Updated; The invocation metadata is expected to be a tuple.

andrewsg (approved):

tritone (approved):
LGTM so far, might have more feedback once you update the readme :) Thanks Cathy!

PR #533, opened 2023-09-05
feat: add retry test inject failure midway in gRPC ReadObject

cojenco:
-  Add Retry Test API support for injecting failure midway in gRPC ReadObject
    - Testbench will fail with UNAVAILABLE after a few downloaded data with retry instructions return-broken-stream
    - Testbench will fail with UNAVAILABLE after YKiB of downloaded data with retry instructions return-broken-stream-after-YK
- Update README
- Enable methods that have Retry Test support in gRPC
    -       "storage.buckets.get",
            "storage.buckets.getIamPolicy",
            "storage.buckets.list",
            "storage.hmacKey.get",
            "storage.hmacKey.list",
            "storage.notifications.get",
            "storage.notifications.list",
            "storage.objects.list",
            "storage.objects.get",
            "storage.serviceaccount.get",

- [x] Tests pass


README.md
```diff
@@ -209,12 +209,15 @@ and to some degree decouples the test setup from the test execution.
...
+when a buckets list operation is received by the testbench JSON server with the returned
 retry test ID.
 
+Note: A Retry Test resource can only apply to one transport, either JSON or GRPC. It defaults
+to "JSON" if not specified.
+
 ```bash
 curl -X POST "http://localhost:9000/retry_test" -H 'Content-Type: application/json' \
-     -d '{"instructions":{"storage.buckets.list": ["return-503"]}}'
+     -d '{"instructions":{"storage.buckets.list": ["return-503"]}, "transport": "JSON"}'
```
tritone, line 220: IMO the options should really be HTTP or gRPC. The server does support (a few) XML endpoints for the retry test API currently and I assume those would still be supported.

README.md
```diff
@@ -236,9 +239,13 @@ curl -X DELETE "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317
...
 
 The following cURL request will attempt to list buckets and the testbench will emit
-a `503` error once based on the Retry Test created above. Subsequent list buckets
+a `503` error once in the JSON server based on the Retry Test created above. Subsequent list buckets
 operations will succeed.
 
+In this example, list buckets operations in the gRPC server will not be impacted since the Retry Test
+created above only applies to JSON. Specify `"transport": "GRPC"` for a Retry Test Resource to apply
+to gRPC operations.
+
```
tritone, line 248: So, it looks like the errors/instructions are the same in gRPC vs. JSON. I'm wondering if this is really what we want. Since gRPC has a well definite set of error codes (even a broken stream will be turned into a specific gRPC error in the gRPC client-side code) I'm wondering if we should be accepting these codes in the retry test API rather than mapping from HTTP and/or doing stuff like forcing a broken stream. Thoughts?

tritone (commented):
One high level question about the approach; let me know what you think.

README.md
cojenco: The few XML endpoints are still supported. Happy to update this to HTTP

README.md
cojenco: I'm under the impression that we wanted to use the same retry conformance test schema to run tests under both transports, and have that outlined in the design sketch. One of the biggest pros is that it relieves the refactoring burden on the clients. I think this remains a good starting point for conformance testing (i.e. retry for UNAVAILABLE; does not retry for UNAUTHENTICATED etc)

In the current design and implementation, there is a separate retry handler for the grpc server, so that will give us flexibility. We can further define what failure injections make more sense in the gRPC context. Scheduled some time with you to discuss offline :)

cojenco:
As discussed offline, I updated Retry Test API Forced Failures Support description by transport and clarified how retry instructions result in the gRPC context. @tritone PTAL, thank you!

README.md
```diff
@@ -247,9 +254,9 @@ curl -H "x-retry-test-id: 1d05c20627844214a9ff7cbcf696317d" "http://localhost:91
 
 | Failure Id              | Description
 | ----------------------- | ---
-| return-X                                  | Testbench will fail with HTTP code provided for `X`, e.g. return-503 returns a 503
-| return-X-after-YK                         | Testbench will return X after YKiB of uploaded data
-| return-broken-stream-final-chunk-after-YB | Testbench will break connection on final chunk of a resumable upload after Y bytes
-| return-broken-stream                      | Testbench will fail after a few bytes
-| return-broken-stream-after-YK             | Testbench will fail after YKiB of downloaded data
-| return-reset-connection                   | Testbench will fail with a reset connection
+| return-X                                  | [HTTP] Testbench will fail with HTTP code provided for X, e.g. return-503 returns a 503  <br> [GRPC] Testbench will fail with X that is supported in the gRPC context (`400`, `401`, `429`, `500`, `501`, `503`)
```
tritone, line 257: I would clarify that it will fail with the equivalent gRPC error to the HTTP code.

tritone (approved):
Looks good to me; thanks for the updates.

cojenco:
@coryan Wanted to check with you for future planning if there is any need to support `x-goog-emulator-instructions` force failures (e.g. return-503-after-256K) in the grpc server?

This PR adds support to the Retry Test API. Planning to merge today if no issues, thanks!

coryan:
> @coryan Wanted to check with you for future planning if there is any need to support `x-goog-emulator-instructions` force failures (e.g. return-503-after-256K) in the grpc server?

No, and we can always add the support if it becomes needed.

PR #545, opened 2023-10-04
feat: support pipx

### LINKED ISSUE #544: Add entrypoints for pipx
**Is your feature request related to a problem? Please describe.**

Since storage-testbench is a Python project providing a single executable command, it would be nice if it could be managed using [pipx](https://pypa.github.io/pipx/). It would make it easier to run without having to know which Python was used to install it.

The pipx docs describe how to add entry points to `setup.py` in order to support running with pipx:
https://pypa.github.io/pipx/how-pipx-works/#developing-for-pipx

**Describe alternatives you've considered**

What we're doing right now, for Apache Arrow, is install the testbench in a Python environment that's also used for unrelated things. It creates coupling between different pieces of our CI, and also requires passing/finding the proper Python executable even from non-Python tests, such as in the following C++ code:
https://github.com/apache/arrow/blob/532ab464db39da0e6b03e4b046355e7bf499730a/cpp/src/arrow/filesystem/gcsfs_test.cc#L92-L133

coryan:
Fixes #544

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,4 +53,10 @@ setuptools.setup(
         "gunicorn==21.2.0",
         "waitress==2.1.2",
     ],
+    # Support installing via https://pypa.github.io/pipx/
+    entry_points={
+        "console_scripts": [
+            "storage-testbench = testbench.rest_server:_main",
+        ]
+    },
 )


coryan:
Sadly our builds  are broken at HEAD, even though the last commit in `main` passed:

https://github.com/googleapis/storage-testbench/actions/runs/6356186515

I have asked help from people with more Python expertise than me.

/FYI: @pitrou

pitrou:
IIUC, the error has nothing to do with the `setup.py` changes? It seems about an error message relating to hashing or base64 encoding:
https://github.com/googleapis/storage-testbench/actions/runs/6407286976/job/17393753132?pr=545#step:5:81

pitrou:
It seems there is the same problem on another PR: https://github.com/googleapis/storage-testbench/actions/runs/6390316476, so perhaps some upstream dependency changed slightly and broke things?

coryan:
> IIUC, the error has nothing to do with the `setup.py` changes?

I agree, but I ~am paranoid~ try to be disciplined.  

> so perhaps some upstream dependency changed slightly and broke things?

That seems the most likely explanation.  Though this code only depends on core Python library modules (`base64`, `hashlib`).  They can break, but break across multiple versions of Python?  Both in GHA and on my workstation?  Could be, but it is more likely that some other module monkey patched something (incorrectly), or that I do not know how Python dependency management works.  My vote is on the latter.

pitrou:
Ok, here is a successful build:
https://github.com/googleapis/storage-testbench/actions/runs/6356186515/job/17265417749
Here is the same build failing:
https://github.com/googleapis/storage-testbench/actions/runs/6390316476/job/17343327057

I've tried to find the package differences between them:

| Package | Success | Failure |
| - | - | - |
| Werkzeug  | 2.3.7 | 3.0.0 |
| charset-normalizer  | 3.2.0 | 3.3.0 |
| urllib3 | 2.0.5 | 2.0.6 |
| coverage | 7.3.1 | 7.3.2 |
| packaging | 23.1 | 23.2 |

(I would suggest adding a call to `pip list` at the end of your "install dependencies" build step to make this easier :-))

Given how the error looks like, I would suggest looking at the Werkzeug update for a potential culprit, and perhaps urllib3.

coryan:
@pitrou wow! thanks for the detailed analysis, I am sorry you had to do this.  I ran the test with the "success" versions and this may be "a clue":

```
  /usr/local/google/home/coryan/storage-testbench/.venv/lib/python3.11/site-packages/werkzeug/datastructures/headers.py:297: DeprecationWarning: Passing bytes as a header value is deprecated and will not be supported in Werkzeug 3.0.
    _value = _str_header_value(_value)
```

I am going to send separate PRs to fix the problem, and to pin these indirect dependencies. The latter should help us detect breakages when they are updated.

tritone (approved):

pitrou:
@jorisvandenbossche @raulcd This should make it easier to dissociate the Python version under test from the one used to run the GCS testbench.

coryan:
FWIW, I just created a "release" (just a tag really).  That may make it easier to use these changes in your CI system.

PR #557, opened 2023-10-12
chore: reduce renovate noise

coryan:
Update docker image only once a week.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -6,7 +6,7 @@
     {
       "matchManagers": ["dockerfile"],
       "matchPackageNames": ["python"],
-      "matchCurrentVersion": "3.11",
+      "matchCurrentVersion": "3.12",
       "schedule": ["before 12pm on Monday"]
     }
   ]


renovate.json
```diff
@@ -6,7 +6,7 @@
     {
       "matchManagers": ["dockerfile"],
       "matchPackageNames": ["python"],
-      "matchCurrentVersion": "3.11",
+      "matchCurrentVersion": "3.12",
```
devbww, line 9: I'm not sure how this matches the "Update docker image only once a week" description.

devbww (approved):

renovate.json
coryan: Strangely enough that is the effect.  This rule says "Update any Python:3.12 Dockerfiles once a weekly" and the Dockerfile we use happens to be 3.12:

https://github.com/googleapis/storage-testbench/blob/5551644fd18f8d8c1d715852f964ff153bdb3dce/Dockerfile#L14

PR #572, opened 2023-11-06
fix: races in Object metadata updates

laramiel:
Tensorstore has a test which does concurrent reads/writes to storage testbench.

(https://github.com/google/tensorstore/blob/41812f367837989d91c7e1ae8548ecfa166e70be/tensorstore/kvstore/gcs_grpc/gcs_grpc_testbench_test.cc#L169)

The test spawns N threads. Each thread uses conditional operations to atomically update a region of the object.

This test currently is flaky due to missing object locking in storage-testbench.

Before this change, the database returned an object which was modified outside of the lock.
After this change, the modifications to the database object happen while a lock is held.

### DIFF
--- a/testbench/database.py
+++ b/testbench/database.py
@@ -18,9 +18,14 @@ import os
 import pathlib
 import threading
 import uuid
+import copy
+from typing import TypeVar, Any, Callable
 
 import gcs
 import testbench
+
+
+T = TypeVar('T')
 
 
 class Database:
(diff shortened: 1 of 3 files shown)


testbench/database.py
```diff
@@ -332,6 +338,22 @@ def delete_object(
             bucket = self.__get_bucket_for_object(bucket_name, context)
             bucket.pop("%s#%d" % (blob.metadata.name, blob.metadata.generation), None)
 
+    def do_update(
```
coryan, line 341: Please rename this to `do_object_update` or something similar.  There are other "things" in the database and it is unclear what this will act upon.

coryan (changes requested):

testbench/database.py
```diff
@@ -175,9 +179,6 @@ def __extract_list_object_request_grpc(cls, request):
     def __extract_list_object_request(cls, request, context):
         if context is not None:
             return cls.__extract_list_object_request_grpc(request)
-        delimiter, prefix, versions = "", "", False
```
coryan, line 178: BTW thanks for removing this redundant code.

coryan (approved):

PR #573, opened 2023-11-14
fix: handle WriteObject final message with empty data

### LINKED ISSUE #574: grpc: object upload issues found by java retry conformance tests
While updating the java retry conformance suite to also run for `transport: GRPC` I ran into two oddities in the testbench `retry_test` integration with grpc.

### `storage.object.create` instructions do not apply to `StartResumableWrite`

Java treats all resumable upload PUTs as idempotent. Because the instruction isn't being processed until `WriteObject`, java still retries. This is inconsistent with JSONs handling, which does apply instructions to starting a resumable upload.

### `WriteObject` does not finalize a resumable session sometimes

When testbench receives a `WriteObjectRequest` with `finish_write: true` without any `checksummed_data` it incorrectly response with an incremental response rather than the resources.

* `>>>` message sent
* `<<<` message received

```
>>> {
  upload_id: "ef1d103dc03cc220587514ab957cb56e3d9e80af394af3b5d05f8ba97602a94f"
  checksummed_data {
    content: "AAAAAAAAAAAAAAAAAAAA<snip (262144)>"
  }
}
<<< {
  persisted_size: 262144
}
## resumable session left open for further appends

## EOF signaled in app, finalize resumable session
>>> {
  upload_id: "ef1d103dc03cc220587514ab957cb56e3d9e80af394af3b5d05f8ba97602a94f"
  write_offset: 262144
  finish_write: true
}
<<< {
  persisted_size: 262144
}
```

The `checksummed_data` in `WriteObjectRequest` is in the [`oneof data`](https://github.com/googleapis/googleapis/blob/2428362137f0833a0d98c797b1b0a4a0f940af62/google/storage/v2/storage.proto#L1075-L1080) declaration. Because it is declared in the `oneof` it is implicitly optional and need to be present in the message. 

[AIP-146](https://google.aip.dev/146) touches on this briefly
> Because the individual fields in the oneof have different keys, a developer can programmatically determine which (if any) of the fields is populated.

Testbench will need to be updated to account for this.

cojenco:
- Return object resource and finalize a resumable session for a `WriteObjectRequest` w/ `finish_write: true` w/o checksummed_data
- Apply grpc retry wrapper to `StartResumableWrite`
- Updated tests

Fixes #574

### DIFF
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -201,6 +201,10 @@ class Upload(types.SimpleNamespace):
             data = request.WhichOneof("data")
             if data == "checksummed_data":
                 checksummed_data = request.checksummed_data
+            elif data is None and request.finish_write:
+                # Handles final message with no data to insert.
+                upload.complete = True
+                continue
             else:
                 print("WARNING unexpected data field %s\n" % data)
                 continue
--- a/testbench/grpc_server.py
+++ b/testbench/grpc_server.py
@@ -776,6 +776,7 @@ class StorageServicer(storage_pb2_grpc.StorageServicer):
 
         return response
 
+    @retry_test(method="storage.objects.insert")
     def StartResumableWrite(self, request, context):
         bucket = self.__get_bucket(request.write_object_spec.resource.bucket, context)
         upload = gcs.upload.Upload.init_resumable_grpc(request, bucket, context)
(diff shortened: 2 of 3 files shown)


BenWhitehead (approved):
Thanks so much for these fixes @cojenco!

coryan (approved):

PR #586, opened 2024-01-24
feat(gRPC): implement BidiWriteObject

### LINKED ISSUE #568: Implement bidirectional writes in GRPC

cojenco:
Fixes #568

### DIFF
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -296,7 +296,9 @@ class Upload(types.SimpleNamespace):
                 if upload.complete:
                     # Resuming an already finalized object will result with a response
                     # containing the uploaded object's metadata.
-                    yield storage_pb2.BidiWriteObjectResponse(resource=upload.blob.metadata)
+                    yield storage_pb2.BidiWriteObjectResponse(
+                        resource=upload.blob.metadata
+                    )
                 is_resumable = True
             elif (
                 first_message == "write_object_spec"
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -294,9 +294,9 @@ class Upload(types.SimpleNamespace):
             if first_message == "upload_id":  # resumable upload
                 upload = db.get_upload(request.upload_id, context)
                 if upload.complete:
-                    return testbench.error.invalid(
-                        "Uploading to a completed upload %s" % upload.upload_id, context
-                    )
+                    # Resuming an already finalized object will result with a response
+                    # containing the uploaded object's metadata.
+                    yield storage_pb2.BidiWriteObjectResponse(resource=upload.blob.metadata)
                 is_resumable = True
             elif (
                 first_message == "write_object_spec"
@@ -331,11 +331,7 @@ class Upload(types.SimpleNamespace):
                 # Handles final message with no data to insert.
                 upload.complete = True
                 continue
-            else:
-                return testbench.error.invalid(
-                    "Invalid data field, should be one of checksummed_data",
-                    context,
-                )
+
             content = checksummed_data.content
             crc32c_hash = (
                 checksummed_data.crc32c if checksummed_data.HasField("crc32c") else None
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -286,9 +286,6 @@ class Upload(types.SimpleNamespace):
     def process_bidi_write_object_grpc(cls, db, request_iterator, context):
         """Process a BidiWriteObject streaming RPC, and yield a stream of responses."""
         upload, object_checksums, is_resumable = None, None, False
-        # Currently, the testbench will checkpoint and flush the data to upload.media per request for testing purposes,
-        # instead of the 15 seconds interval used in the GCS server.
-        should_checkpoint_session = True
         for request in request_iterator:
             first_message = request.WhichOneof("first_message")
             if first_message == "upload_id":  # resumable upload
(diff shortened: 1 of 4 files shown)


gcs/upload.py
```diff
@@ -280,6 +282,137 @@ def init_write_object_grpc(cls, db, request_iterator, context):
             upload.metadata.metadata["x_emulator_no_md5"] = "true"
         return upload, is_resumable
 
+    @classmethod
+    def process_bidi_write_object_grpc(cls, db, request_iterator, context):
```
coryan, line 286: optional: this is a rather long function, consider breaking it down, maybe split up the function to process the first message?

tests/test_upload.py
```diff
@@ -694,6 +694,300 @@ def test_init_object_write_grpc_final_message_empty_data(self):
         self.assertEqual(upload.metadata.name, "object")
         self.assertEqual(upload.metadata.bucket, "projects/_/buckets/bucket-name")
 
+    def test_process_bidi_write_grpc_cannot_resume_completed_upload(self):
```
coryan, line 697: Hmm... why?  Consider this scenario:

- The client is trying to resume an upload
- The client sends a QueryResumableUpload() RPC and gets a result of saying that the upload is not completed.
- Some other client (or a delayed message in the service internal processes) completes the upload
- The client tries to resume the upload by sending the same data as before, and finalizing.

I would think that should succeed with the previously uploaded object.

tests/test_upload.py
```diff
@@ -694,6 +694,300 @@ def test_init_object_write_grpc_final_message_empty_data(self):
...
+        db = unittest.mock.Mock()
+        db.get_bucket = unittest.mock.MagicMock(return_value=bucket)
+        db.get_upload = unittest.mock.MagicMock(return_value=upload)
+        responses = list(
+            gcs.upload.Upload.process_bidi_write_object_grpc(db, [r1, r2], context)
+        )
+        blob = responses[0].resource
+        self.assertEqual(blob.name, "object")
+        self.assertEqual(blob.bucket, "projects/_/buckets/bucket-name")
+
```
coryan, line 990: Consider a test for "invalid data field":

https://app.codecov.io/gh/googleapis/storage-testbench/pull/586?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=googleapis#83ee5f7e84364b0abad9908f111e7409-R335

coryan (commented):
A couple of optional things and a question.

tests/test_upload.py
coryan: I see this is fixed.  FWIW, don't feel that you need to wait for another review from me.

tests/test_upload.py
cojenco: On second thought, we probably don't need this invalid field check after all. I think in most scenarios we can rely on protobuf message checking - for a oneof group, checks if any field inside is set and raises ValueError for anything not defined in the message descriptor.

gcs/upload.py
cojenco: Agreed, I'm still trying to see how this can be split up without affecting the generator

gcs/upload.py
```diff
@@ -280,6 +282,135 @@ def init_write_object_grpc(cls, db, request_iterator, context):
...
+                        "crc32c in checksummed data",
+                        crc32c_hash,
+                        actual_crc32c,
+                        context,
+                    )
+
+            # The testbench should ignore any request bytes that have already been persisted,
+            # thus we validate write_offset against persisted_size.
+            # https://github.com/googleapis/googleapis/blob/15b48f9ed0ae8b034e753c6895eb045f436e257c/google/storage/v2/storage.proto#L320-L329
+            if request.flush or should_checkpoint_session:
```
coryan, line 354: This is a complicated way to write `if True` since `should_checkpoint_session` is always `True`?

Consider rewriting the code to just always checkpoint and then leave a TODO to implement behavior that more closely follows what the service does.

gcs/upload.py
```diff
@@ -280,6 +282,135 @@ def init_write_object_grpc(cls, db, request_iterator, context):
             upload.metadata.metadata["x_emulator_no_md5"] = "true"
         return upload, is_resumable
 
+    @classmethod
+    def process_bidi_write_object_grpc(cls, db, request_iterator, context):
+        """Process a BidiWriteObject streaming RPC, and yield a stream of responses."""
+        upload, object_checksums, is_resumable = None, None, False
+        # Currently, the testbench will checkpoint and flush the data to upload.media per request for testing purposes,
+        # instead of the 15 seconds interval used in the GCS server.
+        should_checkpoint_session = True
```
coryan, line 291: Just an idea (no action needed): maybe we should let the client pick the behavior (via headers like we do for retry testing). That way can write tests for different scenarios (e.g. when the data is not flushed and we disconnect, or when the data is flushed before the client expects it to).

gcs/upload.py
```diff
@@ -280,6 +282,135 @@ def init_write_object_grpc(cls, db, request_iterator, context):
...
+                    )
+                upload.media += content
+            if request.finish_write:
+                upload.complete = True
+            if request.state_lookup:
+                yield storage_pb2.BidiWriteObjectResponse(
+                    persisted_size=len(upload.media)
+                )
+
+        if upload is None:
```
coryan, line 368: Checking this after doing `upload.media += content` seems like a bug.

Seems like this check should be around line 350 or so?

gcs/upload.py
```diff
@@ -280,6 +282,135 @@ def init_write_object_grpc(cls, db, request_iterator, context):
...
+            if request.flush or should_checkpoint_session:
+                if request.write_offset < len(upload.media):
+                    range_start = len(upload.media) - request.write_offset
+                    content = testbench.common.partial_media(
+                        content, range_end=len(content), range_start=range_start
+                    )
+                upload.media += content
+            if request.finish_write:
+                upload.complete = True
+            if request.state_lookup:
```
coryan, line 363: I think if `request.finish_write == True and request.state_lookup == True` you do not want to yield just the persisted bytes?

coryan (changes requested):

gcs/upload.py
cojenco: haha yes it's quite the illusion xD 
Updated with comments and linked to a TODO issue

gcs/upload.py
cojenco: This is to handle cases with missing BidiWriteObjectRequest. I probably went overboard, and there's also a test case for this `gcs.upload.Upload.process_bidi_write_object_grpc(db, [], context)`
https://github.com/googleapis/storage-testbench/blob/65105b6fdf5efcf590326bac65650076b3677824/tests/test_upload.py#L646-L653

gcs/upload.py
cojenco: Good catch. I've updated this part of the logic so a response with persisted_size is only yielded for **incomplete uploads**. Definitely not elegant... my thought process here is for completed uploads when `finish_write = True` to be finalized out of the loop (store full object checksums, create new object, insert to database, and yield final response with resource metadata)... thoughts?

gcs/upload.py
coryan: Right, but if `upload` is None then line 359 would have raised an exception before you got here, right?   You probably want to detect the problem before the first use of `upload`.  Did I miss something?  If so, please just merge, otherwise please reorder the test to prevent ~crashes~ exceptions.

gcs/upload.py
coryan: SGTM

coryan (approved):

gcs/upload.py
cojenco: Inside the request loop on L307, we're checking if `upload` is None before the first use. 

This extra check is outside the request loop. It covers the rare case of missing requests, and prevents crashes in the attempt to finalize an upload

PR #593, opened 2024-02-15
feat: support bucket soft-delete policies

coryan:
Part of the work for https://github.com/googleapis/google-cloud-cpp/issues/13575

### DIFF
--- a/gcs/bucket.py
+++ b/gcs/bucket.py
@@ -44,6 +44,7 @@ class Bucket:
         "encryption",
         "billing",
         "retention_policy",
+        "soft_delete_policy",
         "location_type",
         "iam_config",
         "rpo",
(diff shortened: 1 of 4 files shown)


gcs/bucket.py
```diff
@@ -285,6 +309,26 @@ def _init_defaults(cls, metadata: storage_pb2.Bucket, context):
...
+                "SoftDeletePolicy.retention_duration should not have nanoseconds",
+                context,
+            )
+        if policy.retention_duration.ToSeconds() < 7 * 86400:
+            testbench.error.invalid(
+                "SoftDeletePolicy.retention_duration should be at least 7 days", context
+            )
+        if policy.retention_duration.ToSeconds() >= 365 * 86400:
+            testbench.error.invalid(
+                "SoftDeletePolicy.retention_duration should be at less than a year",
```
cojenco, line 328: Is soft delete retention_duration maximum 90 days? This is what I see in the user guide:
`The value must be greater than or equal to 604,800 seconds (7 days) and less than 7,776,000 seconds (90 days)`

nit: extra at within error message

cojenco (approved):

PR #598, opened 2024-02-24
test: add mutation test case

cojenco:
Context: mutation test caught missing test case in internal cl/609739391

Not sure why productive coverage check is angry as there is test coverage through
https://github.com/googleapis/storage-testbench/blob/a60a7145a67098e7c6cd4142391e240b3d462ecc/tests/test_grpc_server.py#L2113-L2151

### DIFF
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -208,8 +208,8 @@ class Upload(types.SimpleNamespace):
                 upload.complete = True
                 continue
             else:
-                print("WARNING unexpected data field %s\n" % data)
-                continue
+                testbench.error.invalid("Invalid data field in upload", context)
+                return None, False
 
             content = checksummed_data.content
             crc32c_hash = (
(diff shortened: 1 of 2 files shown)


gcs/upload.py
```diff
@@ -330,6 +331,9 @@ def process_bidi_write_object_grpc(cls, db, request_iterator, context):
                 # Handles final message with no data to insert.
                 upload.complete = True
                 continue
+            else:
+                print("WARNING unexpected data field %s\n" % data)
```
coryan, line 335: Is this a debugging remnant?  We don't use `print()` anywhere else in the code.  Should we return an error to the client?

coryan (approved):

gcs/upload.py
cojenco: Right, it makes more sense to return an error. Updated both WriteObject and BidiWriteObject plus tests.

coryan (approved):

PR #608, opened 2024-03-08
chore: avoid forbidden terms

coryan:


cojenco (approved):
Thanks for the change!

PR #619, opened 2024-03-20
feat: support retry test api in BidiWriteObject

cojenco:
This adds retry test api support in gRPC BidiWriteObject

### DIFF
--- a/tests/test_testbench_retry.py
+++ b/tests/test_testbench_retry.py
@@ -795,6 +795,7 @@ class TestTestbenchRetryGrpc(unittest.TestCase):
                 {
                     "instructions": {
                         "storage.objects.insert": [
+                            "return-503",
                             "return-503-after-256K",
                             "return-503-after-300K",
                         ]
@@ -822,6 +823,7 @@ class TestTestbenchRetryGrpc(unittest.TestCase):
             ),
             context=context,
         )
+        context.abort.assert_called_with(StatusCode.UNAVAILABLE, unittest.mock.ANY)
         self.assertIsNotNone(start.upload_id)
 
         # Upload the first 256KiB chunk of data and trigger error.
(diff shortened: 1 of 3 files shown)


tests/test_testbench_retry.py
```diff
@@ -786,6 +786,78 @@ def test_grpc_return_error_after_bytes(self):
         self.assertEqual(blob.bucket, "projects/_/buckets/bucket-name")
         self.assertEqual(blob.size, 2 * UPLOAD_QUANTUM)
 
+    def test_grpc_bidiwrite_return_error_after_bytes(self):
+        # Setup two after-bytes errors to test injecting failures in
```
tritone, line 790: Do we need a test for initial-response errors as well? Or are existing tests sufficient for this?

tests/test_testbench_retry.py
```diff
@@ -786,6 +786,78 @@ def test_grpc_return_error_after_bytes(self):
...
+            ),
+            finish_write=True,
+        )
+        streamer = self.grpc.BidiWriteObject([r2], context)
+        responses = list(streamer)
+        context.abort.assert_called_with(StatusCode.UNAVAILABLE, unittest.mock.ANY)
+        blob = responses[0].resource
+        self.assertEqual(blob.name, "object-name")
+        self.assertEqual(blob.bucket, "projects/_/buckets/bucket-name")
+        self.assertEqual(blob.size, 2 * UPLOAD_QUANTUM)
```
tritone, line 859: Should this also include checking that the upload succeeds after the errors have been fully used up?

tritone (approved):
Looks good, couple minor questions re: testing.

tests/test_testbench_retry.py
cojenco: Thanks for the review! Yep added a check that the upload succeeds with the accurate size

tests/test_testbench_retry.py
cojenco: There are existing tests for initial-response errors. It doesn't hurt to add an extra check here, updated! Will be merging shortly

tritone (approved):

PR #622, opened 2024-03-23
fix: send correct content-range header if requested byte range was truncated

jbms:
Previously, if the requested byte range was truncated, an invalid content-range header was sent.


coryan (approved):
Thanks for the PR!

PR #628, opened 2024-04-11
feat: enable more methods for retry test support in gRPC

cojenco:
Allow more methods for Retry Test support in gRPC

- equip gRPC endpoints with @retry_test decorator
- update `__validate_grpc_method_implemented_retry` guard
- 501 Unimplemented exceptions are only raised for ACL and notifications test cases that are not applicable in gRPC

### DIFF
--- a/testbench/database.py
+++ b/testbench/database.py
@@ -446,22 +446,29 @@ class Database:
     def __validate_grpc_method_implemented_retry(self, method):
         """Returns Unimplemented 501 for methods that are not yet supported.
         Temporary validation while adding Retry Test API support in gRPC."""
-        implemented_grpc_w_retry = {
-            "storage.buckets.get",
-            "storage.buckets.getIamPolicy",
-            "storage.buckets.list",
-            "storage.hmacKey.get",
-            "storage.hmacKey.list",
-            "storage.notifications.get",
-            "storage.notifications.list",
-            "storage.objects.insert",
-            "storage.objects.list",
-            "storage.objects.get",
-            "storage.serviceaccount.get",
+        not_supported_grpc_w_retry = {
+            "storage.bucket_acl.get",
+            "storage.bucket_acl.list",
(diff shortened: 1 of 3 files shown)


testbench/database.py
```diff
@@ -446,22 +446,30 @@ def __validate_injected_failure_description(self, failure):
...
+            "storage.default_object_acl.delete",
+            "storage.default_object_acl.insert",
+            "storage.default_object_acl.patch",
+            "storage.default_object_acl.update",
+            "storage.object_acl.get",
+            "storage.object_acl.list",
+            "storage.object_acl.delete",
+            "storage.object_acl.insert",
+            "storage.object_acl.patch",
+            "storage.object_acl.update",
```
tritone, line 467: I think notifications should also be skipped since it's not implemented server-side for storage/v2.

tritone (approved):
I tested the PR out for Go and the only thing that fails is notifications methods in gRPC. Once we skip those I'm good to merge.

testbench/database.py
cojenco: Got it, thanks! Updated to skip notifications test cases

coryan: > I think notifications should also be skipped since it's not implemented server-side for storage/v2.

But they will be implemented in storage/v2, right?

tritone: They will only be in storage control.

tritone (approved):

testbench/database.py
tritone: Actually @coryan informed me that this is incorrect; the plan of record is to add them to storage/v2 as well.

cojenco:
Checked with java folks - no issues with merging while we're tracking work in internal go/gcs-testbench-grpc-server-client

Codecov upload issue is irrelevant to the changes in this PR; #630 and internal b/334987416 track next steps

Issue #630, opened 2024-04-15
CI: codecov upload fails


cojenco:
Add in a codecov upload token might provide a remedy for failed codecov uploads, based on https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954 and https://github.com/codecov/codecov-action/issues/598

I could not access https://app.codecov.io/gh/googleapis/storage-testbench/settings nor retrieve a codecov token, filed internal b/334987416 for support

cojenco:
Fixed via #635

PR #631, opened 2024-04-16
chore: bump version post v0.44.0

cojenco:


coryan:
The code coverage uploads seem to be broken. But that is no reason to block this PR.

cojenco:
> The code coverage uploads seem to be broken. But that is no reason to block this PR.

I think we need to add a codecov upload token. Filed #630 and internal b/334987416 to obtain a token.

PR #684, opened 2024-10-10
feat: add write stall support 

Tulsishah:
Fixes https://github.com/googleapis/storage-testbench/issues/686

- [x] Tests pass


tests/test_testbench_retry.py
```diff
@@ -655,6 +655,107 @@ def test_retry_test_return_error_after_bytes(self):
...
+                "x-retry-test-id": test_id,
+            },
+            data=chunk,
+        )
+        end_time = time.perf_counter()
+        elapsed_time = end_time - start_time
+        self.assertGreater(elapsed_time, 1)
+        self.assertEqual(response.status_code, 200, msg=response.data)
+
+        # Upload the second 256KiB chunk of data and check that stall not happen
```
frankyn, line 740: This is essentially a replay which should return a 200 again, is that right?

tests/test_testbench_retry.py
```diff
@@ -655,6 +655,107 @@ def test_retry_test_return_error_after_bytes(self):
...
+        )
+        self.assertEqual(response.status_code, 200)
+
+        location = response.headers.get("location")
+        self.assertIn("upload_id=", location)
+        match = re.search(r"[&?]upload_id=([^&]+)", location)
+        self.assertIsNotNone(match, msg=location)
+        upload_id = match.group(1)
+
+        # Upload the first 256KiB chunk of data and trigger the stall.
```
frankyn, line 699: The only test I don't see is stall with incremental progress from a multiple of quantum

1. Start upload
2. Upload `UPLOAD_QUANTUM * 2` stall after first quantum; assuming you would end connection in retries and query service for offset
3. Retry upload from offset `UPLOAD_QUANTUM-1`

testbench/rest_server.py
```diff
@@ -1104,6 +1104,26 @@ def resumable_upload_chunk(bucket_name):
...
+                stall_time = 10
+                after_bytes = 262144
+            testbench.common.handle_stall_uploads_after_bytes(
+                upload,
+                data,
+                db,
+                stall_time,
+                after_bytes,
+                test_id,
+            )
```
frankyn, line 1126: No concerns here since it follows existing pattern.

testbench/common.py
```diff
@@ -844,24 +843,24 @@ def handle_retry_test_instruction(database, request, socket_closer, method):
...
         return __get_stream_and_stall_fn(
             database, method, test_id, limit=after_bytes, stall_time_sec=stall_time
         )
 
-    if broken_stream_after_bytes and method == "storage.objects.get":
-        items = list(broken_stream_after_bytes.groups())
-        after_bytes = int(items[0]) * 1024
-        return __get_streamer_response_fn(
-            database, method, socket_closer, test_id, limit=after_bytes
-        )
```
frankyn, line 864: Is this code block move necessary?

testbench/common.py
```diff
@@ -844,24 +843,24 @@ def handle_retry_test_instruction(database, request, socket_closer, method):
...
+        after_bytes = int(items[0]) * 1024
+        return __get_streamer_response_fn(
+            database, method, socket_closer, test_id, limit=after_bytes
+        )
 
     retry_stall_after_bytes_matches = testbench.common.retry_stall_after_bytes.match(
         next_instruction
     )
-    if retry_stall_after_bytes_matches:
+    if retry_stall_after_bytes_matches and method == "storage.objects.get":
```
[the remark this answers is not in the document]
frankyn: Ignore this ask; from offline chat I misunderstood that this isn't a bug and more it's possible for this condition to match write stall support when it's not in fact the same; this additional condition check makes this check specific only to stall downloads.

testbench/common.py
Tulsishah: I could have ignored it, but since `broken_stream_after_bytes` was initialized above and used after `retry_stall_after_bytes_matches`  block, it seemed incorrect to me. So, I fixed it here.

tests/test_testbench_retry.py
Tulsishah: Yes

testbench/rest_server.py
frankyn: No action required; it's a note for me mainly.

tests/test_testbench_retry.py
frankyn: This is scoped for write stall support under `stall-xs-after-yk` but outside of scope for this PR; thanks for clarifying @Tulsishah

Tulsishah: As discussed offline this is not in the scope of this PR

frankyn (approved):

testbench/common.py
```diff
@@ -894,6 +893,33 @@ def wrapper(*args, **kwargs):
 
     return retry_test
```
tritone, line 895: Could you add this instruction to the README as well? Or I guess if the instruction is the same as the reads one, you can update the description to say that this applies to uploads too.

testbench/rest_server.py
```diff
@@ -1104,6 +1104,26 @@ def resumable_upload_chunk(bucket_name):
...
+
+        instruction = testbench.common.extract_instruction(request, context=None)
+        (
+            stall_time,
+            after_bytes,
+            test_id,
+        ) = testbench.common.get_stall_uploads_after_bytes(db, request)
+
+        if stall_time or instruction == "stall-10s-after-256K":
+            if instruction == "stall-10s-after-256K":
```
tritone, line 1116: Why is this special case needed?

testbench/rest_server.py
```diff
@@ -1104,6 +1104,26 @@ def resumable_upload_chunk(bucket_name):
                 chunk_last_byte,
                 test_id,
             )
+
```
tritone, line 1107: I see that this logic is added specifically for resumable uploads. We need to cover single shot as well for Go. Maybe in the object_insert function above?

tests/test_testbench_retry.py
```diff
@@ -655,6 +655,107 @@ def test_retry_test_return_error_after_bytes(self):
         self.assertIn("size", create_rest)
         self.assertEqual(int(create_rest.get("size")), 2 * UPLOAD_QUANTUM)
 
+    def test_write_retry_test_stall_after_bytes(self):
```
tritone, line 658: Nice test. Can we add one for single shot uploads as well?

tritone (commented):
Few suggestions, generally looking good!

testbench/common.py
Tulsishah: Done

testbench/rest_server.py
Tulsishah: I could see for "return-Xs-after-Yk" is doing the [same](https://github.com/googleapis/storage-testbench/blob/main/testbench/rest_server.py#L1092). So I followed same thing here.

tritone: Yeah that was to support non-retry test API instructions that were used in old tests for some clients, see https://github.com/googleapis/storage-testbench/blob/main/README.md?plain=1#L166 . We don't have anything similar for uploads so no need to do this.

testbench/rest_server.py
Tulsishah: Added support for this also for stalls. Although please note that other error codes are not supporting this feature. If you want I can add that support in further PRs.

tests/test_testbench_retry.py
Tulsishah: Added

Tulsishah:
> Few suggestions, generally looking good!

Done. PTAL

tritone (approved):
looks good!

tests/test_testbench_retry.py
```diff
@@ -757,6 +762,73 @@ def test_write_retry_test_stall_after_bytes(self):
         self.assertLess(elapsed_time, 1)
         self.assertEqual(response.status_code, 200, msg=response.data)
 
+    def test_write_retry_test_stall_for_full_uploads(self):
```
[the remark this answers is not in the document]
Tulsishah: Done

testbench/rest_server.py
Tulsishah: I did experiment on `return-503-after-3072K`. It was not throwing any googleapi error until I was uploading data less than chunk size(single shot upload). may be we can discuss offline about this.

tritone (approved):

Issue #694, opened 2024-10-25
return error codes are not working for single shot uploads


Tulsishah:
Thanks for stopping by to let us know something could be better!

I was trying to run the following retry_test:
```
curl -X POST "http://localhost:5000/retry_test" -H 'Content-Type: application/json' -d '{"instructions":{"storage.objects.insert": ["return-503-after-3072K"]},"resources": ["BUCKET", "OBJECT"]}'
```
I observed that the error only occurred when uploading data larger than the chunk size (e.g., during resumable uploads). When uploading data smaller than the chunk size, the error did not happen.

PR #699, opened 2024-10-27
feat: Support error code for single shot upload

Tulsishah:
Fixes https://github.com/googleapis/storage-testbench/issues/694

- [x] Tests pass
- [ ] Appropriate changes to README are included in PR


testbench/rest_server.py
```diff
@@ -979,6 +979,22 @@ def object_insert(bucket_name):
         blob, projection = gcs_type.object.Object.init_media(flask.request, bucket)
     elif upload_type == "multipart":
         blob, projection = gcs_type.object.Object.init_multipart(flask.request, bucket)
+        testbench.common.extract_instruction(request, context=None)
```
tritone, line 982: This should work for both multipart and media; they are just diff flavors of single shot upload (though the Go SDK only uses multipart, but other SDKs might use media). See https://cloud.google.com/storage/docs/json_api/v1/objects/insert

testbench/rest_server.py
```diff
@@ -979,6 +979,22 @@ def object_insert(bucket_name):
...
+        (
+            error_code,
+            after_bytes,
+            test_id,
+        ) = testbench.common.get_retry_uploads_error_after_bytes(db, request)
+        if error_code and len(blob.media) >= after_bytes:
+            if test_id:
+                db.dequeue_next_instruction(test_id, "storage.objects.insert")
+            testbench.error.generic(
+                "Fault injected during a resumable upload",
```
tritone, line 992: Message should say single-shot upload here.

tritone (commented):
Nice addition. We should look into making sure that this case is covered in the Go retry test scenarios as well.

testbench/rest_server.py
Tulsishah: Ohh. Thanks. Done. Supported for both.

testbench/rest_server.py
Tulsishah: Done

testbench/rest_server.py
```diff
@@ -979,6 +979,8 @@ def object_insert(bucket_name):
         blob, projection = gcs_type.object.Object.init_media(flask.request, bucket)
     elif upload_type == "multipart":
         blob, projection = gcs_type.object.Object.init_multipart(flask.request, bucket)
+
+    if upload_type == "media" or upload_type == "multipart":
```
tritone, line 983: I think you can remove this if -- the `if resumable` block above has a return so all remaining uploads are either media or multipart.

tritone (approved):
One more comment, otherwise LGTM

testbench/rest_server.py
Tulsishah: Done.

PR #708, opened 2025-02-24
feat(read): add redirect semantics for BidiReadObject

shubham-diwakar:
Fixes Redirection in case of BidiReadObject. 

This adds instructions for returning a redirect error which includes the read_handle in details.

Other redirections( without read_handle ) are not possible in case of BidiRead.

- [x] Tests pass


cojenco (approved):
Thanks for adding this!

Issue #720, opened 2025-04-10
Add support for unfinalized appendable objects


cojenco:
To support unfinalized appendable objects in the testbench, the blob (metadata size, media, etc) needs to be updated in the database within `Upload.process_bidi_write_object_grpc`, which mainly handles the bidi write business logic.

Below is an example of how `blob.media` and `blob.metadata.size` could be updated to the database. This would support test cases where size or data needs to read back for unfinalized appendable objects. 

Testbench implementaton details should be decided based on the internal design doc and needed test cases.

```
upload.media += content
# ...
                
# Update appendable blob size and media here, as an example.
# TODO: (1) Update_time, size and crc are updated in object metadata.
# TODO: (2) Whether the testbench checks for flush or/and performs a background force-close.
if is_appendable:
    def update_appendable_blob(blob, unused_generation):
        blob.media = upload.media
        blob.metadata.size = len(upload.media)
        return blob

    blob = db.do_update_object(
        upload.bucket.name,
        upload.metadata.name,
        update_fn=update_appendable_blob,
        context=context,
        generation=upload.metadata.generation,
    )
``` 

https://github.com/googleapis/storage-testbench/blob/9a8ca669c6083f75c20e648e0a85392dd8be8e2b/gcs/upload.py#L504-L507

tritone:
One more note to add to this: we would expect a GetObject call to succeed once the write stream for an object has been flushed at least once -- even if the write stream is still open.

Currently when I do this, the call hangs.

Does this seem like something that can be fixed?

PR #721, opened 2025-04-15
feat: update appendable blob size and media in unfinalized writes

cojenco:
Towards #720


tritone (approved):
LGTM. I think we'd need to verify/test separately that a takeover append works.

PR #722, opened 2025-04-23
feat: update appendable blob metadata crc32c field

cojenco:
Updates `blob.metadata.checksums.crc32c` for an appendable blob

Towards #720 and related to #721

### DIFF
--- a/gcs/upload.py
+++ b/gcs/upload.py
@@ -330,6 +330,7 @@ class Upload(types.SimpleNamespace):
             require_live_current_generation=False,
         )
 
+
     @classmethod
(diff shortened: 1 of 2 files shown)


gcs/upload.py
```diff
@@ -433,6 +433,27 @@ def response(msg):
                     msg.resource.size = obj_len
             return msg
 
+        def update_upload_checksums(upload_metadata, object_checksums):
+            if object_checksums is None:
```
tritone, line 437: How would this case arise?

tritone (approved):

PR #724, opened 2025-04-30
fix: Handle partial writes better.

cjc25:
Writes with offsets larger than the media were not handled quite correctly. Address partial writes by first trimming the input content to the correct offset (or returning an error if the write_offset is too long), then handling any persist-then-break retry instructions.

### DIFF
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -1018,24 +1018,17 @@ def handle_grpc_retry_uploads_error_after_bytes(
     database,
     rest_code,
     after_bytes,
-    write_offset,
-    persisted_size,
-    expected_persisted_size,
     test_id,
 ):
     """
     Handle error-after-bytes instructions for resumable uploads in the grpc server and commit only partial data before forcing a testbench error.
     This helper method also ignores request bytes that have already been persisted, which aligns with GCS behavior.
     """
-    if after_bytes > len(upload.media) and after_bytes <= expected_persisted_size:
-        # Ignore request bytes that have already been persisted.
-        range_start = 0
-        if len(upload.media) != 0 and write_offset < persisted_size:
-            range_start = persisted_size - write_offset
+    if after_bytes > len(upload.media) and after_bytes <= len(upload.media) + len(data):
         # Only partial data will be commited due to the instructed interruption.
-        range_end = range_start + (after_bytes - len(upload.media))
+        range_end = after_bytes - len(upload.media)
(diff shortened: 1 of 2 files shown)


cjc25:
> Gonna grab lunch real quick but I think the test is incorrect - it is assuming that we persist more than the error instructions say to persist. Gonna check history first before making that change, though.

We just weren't throwing when errors were injected in the context. This led us to have incorrect expectations in the test, too. Fixed both.

testbench/common.py
```diff
@@ -1018,29 +1018,20 @@ def handle_grpc_retry_uploads_error_after_bytes(
...
-        content = testbench.common.partial_media(
-            data, range_end=range_end, range_start=range_start
-        )
+        range_end = after_bytes - len(upload.media)
+        content = testbench.common.partial_media(data, range_end=range_end)
         upload.media += content
         database.dequeue_next_instruction(test_id, "storage.objects.insert")
         grpc_code = _grpc_forced_failure_from_http_instruction(str(rest_code))
         msg = {"error": {"message": "Retry Test: Caused a {}".format(grpc_code)}}
+        print(f"injecting error {rest_code} {grpc_code}")
```
tritone, line 1034: Remove print log

testbench/common.py
```diff
@@ -1018,29 +1018,20 @@ def handle_grpc_retry_uploads_error_after_bytes(
...
     rest_code,
     after_bytes,
-    write_offset,
-    persisted_size,
-    expected_persisted_size,
     test_id,
 ):
     """
     Handle error-after-bytes instructions for resumable uploads in the grpc server and commit only partial data before forcing a testbench error.
-    This helper method also ignores request bytes that have already been persisted, which aligns with GCS behavior.
```
tritone, line 1028: So will this still be the case after these changes? If so let's keep the comment.

cjc25: This will no longer be the case. I have updated the comment.

We now depend on the caller to do this trimming (I moved it since the caller can do it when it checks the offset before hitting this code).

testbench/common.py
cjc25: Oops! removed

testbench/common.py
tritone: Discussed offline; the testbench will still accept re-sending persisted bytes (as the real service does) but this is removing duplication between layers of the testbench.

tritone (approved):

PR #805, opened 2026-06-29
feat: send crc32c along with persisted size in BidiWriteResponse

krishnamd-jkp:


gcs/upload.py
```diff
@@ -587,6 +587,9 @@ def update_upload_checksums(upload_metadata, object_checksums):
                 # TODO(#592): Refactor testbench checkpointing to more closely follow GCS server behavior.
                 upload.media += content
 
+            persisted_crc32c = crc32c.crc32c(upload.media)
```
v-pratap, line 590: Are we recalculating the CRC32C over the entire upload.media array every time a new chunk is appended?

google/storage/control/v2/storage_control_pb2_grpc.py
v-pratap, line 1: why there are changes in storgae control files?

v-pratap (approved):

gcs/upload.py
krishnamd-jkp: Yeah I think there is no incremental calculation of media as of now in the server. This was already being done in one of the codepath so I just reused the same. We can take the optimization in a different PR

google/storage/control/v2/storage_control_pb2_grpc.py
krishnamd-jkp: These are changes from the script - `./update-protos.sh`

PR #809, opened 2026-07-21
feat: adding read/write stall injection support for gRPC

### LINKED ISSUE #808: Stall support for gRPC Read/Write path
Currently test-bench have support for stall injection using `stall-for-Ts-after-YK` only applicable for HTTP protocol. We need to extend this support for gRPC protocol.

raj-prince:
- Fixes #808
- Also, increased the default value of gRPC server to 8 (as 2 seems too low), given parallel stall test gets intermittently stuck because of stall sleep.

- [X] Tests pass
- [X] Appropriate changes to README are included in PR

### DIFF
--- a/testbench/grpc_server.py
+++ b/testbench/grpc_server.py
@@ -573,7 +573,10 @@ class StorageServicer(storage_pb2_grpc.StorageServicer):
         bucket = self.db.get_bucket(request.destination.bucket, context).metadata
         metadata = storage_pb2.Object()
         metadata.MergeFrom(request.destination)
-        (blob, _,) = gcs.object.Object.init(
+        (
+            blob,
+            _,
+        ) = gcs.object.Object.init(
             request, metadata, composed_media, bucket, True, context
         )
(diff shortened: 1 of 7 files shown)


testbench/grpc_server.py
```diff
@@ -615,12 +650,33 @@ def ReadObject(self, request, context):
             broken_stream_after_bytes = testbench.common.get_broken_stream_after_bytes(
                 next_instruction
             )
+            retry_stall_after_bytes_matches = (
+                testbench.common.retry_stall_after_bytes.match(next_instruction)
+            )
+            if retry_stall_after_bytes_matches:
+                items = list(retry_stall_after_bytes_matches.groups())
+                stall_time = int(items[0])
+                stall_after_bytes = int(items[1]) * 1024
```
cpriti-os, line 659: Coudl we be stalling before the said bytes because of gRPC message size? Here and BidiReadObject

cpriti-os (changes requested):

PR #811, opened 2026-07-22
feat: Add fault injection for finalize and close (half-close)

sahusneha2004:
Fixes b/532527637

Context:
The testbench currently lacks the ability to inject faults exactly when a stream is finalized (finish_write=True) or gracefully half-closed. This feature is required to verify retry logic in the Python SDK's AsyncAppendableObjectWriter during .finalize() and .close() operations (b/532527637).

Changes: 
Added three new instructions to the storage.objects.insert fault injection suite:
return-[error_code]-on-finish-write
return-[error_code]-on-half-close
redirect-send-handle-and-token-[token]-on-finish-write

Implementation:
common.py: Added regex parsers for the new instructions and an http_to_grpc_status converter.
database.py: Registered the new instructions in the validation suite to allow test setup.
upload.py: Added traps inside the BidiWriteObject upload loop. The stream now checks for these instructions and executes context.abort() with the requested error exactly when request.finish_write is received or the stream is half-closed.

### DIFF
--- a/testbench/common.py
+++ b/testbench/common.py
@@ -39,6 +39,10 @@ from google.storage.v2 import storage_pb2
 
 re_remove_index = re.compile(r"\[\d+\]+|^[0-9]+")
 retry_return_error_code = re.compile(r"return-([0-9]+)$")
+retry_return_error_code_on_finish_write = re.compile(
+    r"return-([0-9]+)-on-finish-write$"
+)
+retry_return_error_code_on_half_close = re.compile(r"return-([0-9]+)-on-half-close$")
 retry_return_error_connection = re.compile(r"return-([a-z\-]+)$")
 retry_return_error_after_bytes = re.compile(r"return-([0-9]+)-after-([0-9]+)K$")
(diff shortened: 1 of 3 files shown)


testbench/common.py
```diff
@@ -52,6 +56,9 @@
 retry_return_handle_and_redirection_token = re.compile(
     r"redirect-send-handle-and-token-([a-z\-]+)$"
 )
+retry_return_handle_and_redirection_token_on_finish_write = re.compile(
```
chandra-siri, line 59: I don't think server returns handle / token on finish_write. Can you confirm once? @sahusneha2004

chandra-siri (changes requested):

testbench/common.py
sahusneha2004: A successful finalize response does not contain a write handle. But a redirect error thrown during finalize contain a write handle so the client knows how to retry it.

chandra-siri (approved):

PR #819, opened 2026-09-14
feat: Add gRPC read stall support for Go SDK

cpriti-os:
Fixes #809

- [x] Tests pass

This PR syncs the changes from PR #809 to support gRPC read stalls for the Go SDK, with an additional cleanup of unused variables (`stall_applied`).


testbench/grpc_server.py
```diff
@@ -42,7 +43,38 @@
 from google.storage.control.v2 import storage_control_pb2, storage_control_pb2_grpc
 from google.storage.v2 import storage_pb2, storage_pb2_grpc
 
-_GRPC_SERVER_THREAD_COUNT = 2
+_GRPC_SERVER_THREAD_COUNT = 8
```
krishnamd-jkp, line 46: Why do we need to increase it?

testbench/grpc_server.py
```diff
@@ -42,7 +43,38 @@
...
+):
+    if not test_id:
+        return False
+
+    if stall_time <= 0 or not _should_stall_after_bytes(
+        bytes_yielded, chunk_len, stall_after_bytes
+    ):
+        return False
+
+    if database.dequeue_next_instruction(test_id, method) is None:
```
krishnamd-jkp, line 74: does it not work if stall is the last command then?

krishnamd-jkp (approved):

testbench/grpc_server.py
cpriti-os: This by Prince "Also, increased the default value of gRPC server to 8 (as 2 seems too low), given parallel stall test gets intermittently stuck because of stall sleep."

testbench/grpc_server.py
cpriti-os: yes, if the stall instruction is the very last one in the queue, dequeue_next_instruction will still return that string, which means is None evaluates to False, and the time.sleep(stall_time) executes successfully as expected.

